PR #22, opened 2026-09-06
docs: format Spring benchmark runtime environment

DPS0340:
## Summary

- Split the Spring runtime environment paragraph in `README.md` into runtime details, JVM options, and warmup/profile sections.
- Display each JVM option on its own line in a fenced `text` block.
- Preserve the supplied version strings, runtime configuration, JVM option values, and measurement profile; benchmark results are unchanged.

## Validation

- `git diff --check` passed.
- Reviewed the diff: only `README.md` is changed.
- No build or benchmark run: this is a documentation-formatting-only change, not a revalidation of the reported runtime or performance.

## Scope note

`scripts/ci/benchmark.py` currently generates this section as a single paragraph. A future benchmark refresh may overwrite this formatting. The generator is intentionally unchanged to keep this PR scoped to the requested `README.md` edit.

PR #23, opened 2026-09-07
fix(ci): preserve readable benchmark runtime reports

DPS0340:
## Summary

Fix benchmark report generation so CI preserves readable Spring runtime details. Follow-up to #22.

## Root cause

The `publish-benchmark-report` job in `.github/workflows/bsd-kqueue-benchmarks.yml` runs `python3 scripts/ci/benchmark.py render` and commits the generated documentation. The renderer concatenated the JDK, Spring stack, JVM options, and measurement profile into one paragraph with one long inline-code span. Bot commit `ad2def219dbf024dc7016cb25f7564c42822eec2` overwrote the formatting merged in #22.

## Changes

- Render runtime fields, JVM options, and warmup/profile as separate Markdown sections.
- Put JVM options on separate lines in a fenced text block; preserve quoted values and recorded AOT annotations.
- Render explicit false virtual-thread settings as disabled without inventing missing settings.
- Regenerate both README files from the existing recorded history.
- Add dependency-free CLI regression tests and a read-only Actions test job. Report publication now depends on that job.

No benchmark measurements, history JSON, SVGs, roadmap, or runtime execution settings changed.

## Validation

- Eight unittest regression tests pass. Six fail against the old renderer, confirming regression sensitivity.
- Real-history rendering matches the new README sections and is byte-identical on repeat execution.
- Existing option tokens and throughput lines are preserved.
- GitHub Markdown API confirms all 18 JVM options remain separate lines in a preformatted code block.
- Workflow YAML parsing, publishing-dependency assertions, and `git diff --cached --check` pass.
- Independent staged-diff security and logic review passes with no blocking findings.

Full native builds and fresh performance benchmarks were not run locally; this change only affects report formatting and its tests. Remote Actions results should be checked separately.

### DIFF
--- a/.github/workflows/bsd-kqueue-benchmarks.yml
+++ b/.github/workflows/bsd-kqueue-benchmarks.yml
@@ -15,6 +15,16 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  benchmark-report-tests:
+    name: benchmark report rendering tests
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@v4
+      - name: Test benchmark report generation
+        run: python3 -m unittest discover -s scripts/ci -p 'test_*.py' -v
+
   bsd-kqueue:
     name: kqueue tests (macOS)
     if: github.event_name != 'push' || !contains(github.event.head_commit.message, '[skip ci]')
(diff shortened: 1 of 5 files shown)


gg582:
That's neat. I think I can merge this.

PR #34, opened 2026-09-12
experiment(alloc): mimalloc-backed CWIST build + CI benchmark row (dev only)

gg582:
Follow-up to the allocator discussion in #25. **Update: measured on real CI, mimalloc regresses every metric for this app's C1M/prefork mode — this PR is now the reusable A/B infrastructure and a negative result, not a recommendation.**

## Result (GitHub Actions, 4 vCPU AMD EPYC 7763, same binary/config, allocator as the only variable)

| Metric | CWIST C1M (glibc) | CWIST C1M (mimalloc) |
|---|---|---|
| RPS | 117,431 | 115,279 (-1.8%) |
| P90 | 6.65ms | 6.88ms (+3.5%) |
| P99 | 14.56ms | 16.27ms (+11.7%) |
| P99.999 | 24.18ms | 47.61ms (**+97%**) |
| RSS | 10.3 MB | 22.7 MB (**+120%**) |

**Root cause**: mimalloc eagerly reserves/commits a large arena per process (`arena_reserve=1GiB`, `eager_commit=1` by default) to amortize that cost over a long-lived, allocation-heavy process. CWIST's C1M mode forks one process per core before creating any threads — many short-lived, low-allocation prefork processes, each paying mimalloc's reservation cost with nothing to amortize it against. Not a wiring bug on this branch's part; the override mechanism itself works correctly (verified locally: `malloc` shows up as a local `T` symbol, `MIMALLOC_VERBOSE=1` prints real init/stats per worker) — the allocator's design target just doesn't match this workload.

Full writeup: issue #25.

## What this does
- Vendors mimalloc as a submodule (`lib/mimalloc`, MIT license).
- `USE_MIMALLOC=1` in the top-level `Makefile` builds `libmimalloc.a` via CMake. **Off by default, and staying off** — kept as an opt-in baseline for comparing other allocators/tunings (glibc `mallopt()`, jemalloc with `narenas:1`, ...) against, and in case CWIST's classic thread-pool mode (long-lived, multi-threaded rather than prefork) fits mimalloc's actual design target better — not evaluated here.
[...]

PR #35, opened 2026-09-12
experiment(alloc): cap glibc arena count per worker (CWIST_MALLOC_ARENA_MAX)

gg582:
Follow-up to the mimalloc A/B in #25/#34, which regressed every metric there (RSS +120%, P99.999 +97%) because mimalloc eagerly reserves a large per-process arena to amortize over a long-lived process — a bad fit for CWIST's C1M mode (forks one process per core before creating any threads).

## What this does
Targets the same "N processes × M arenas" multiplication through **glibc's own allocator** instead of swapping in a new one. glibc's arena count is uncapped by default (up to `ncpus*8` on 64-bit) and grows per-process as internal threads contend for the main arena's lock — each C1M worker spins up its own watcher + HTTP/3 threads right after fork, so the same multiplication mimalloc has exists here too, just via thread contention instead of eager reservation, and with zero new dependency or upfront memory cost.

- `src/sys/app/app.c`: `mallopt(M_ARENA_MAX, N)` via `CWIST_MALLOC_ARENA_MAX`, called once before the fork loop in `cwist_app_listen()` so every worker inherits it. `mallopt` state is plain process memory — copied by `fork()` like everything else, no cross-process sharing trick needed (or possible: any inherited page gets private-copied on the child's first write regardless of what it is). **Unset by default: identical behavior to today.** Guarded `#if defined(__GLIBC__)` — `M_ARENA_MAX` is a glibc extension, not available on macOS/BSD malloc.
- CI: adds `CWIST C1M (arena_max=1)` right after `CWIST C1M` — same binary/config, the env var is the only variable, direct A/B. `scripts/ci/benchmark.py` renders it as its own README row.

## Not verified here
Local `wrk` runs on this branch were within this environment's own noise floor — the exact same local machine flip-flopped on which allocator won the mimalloc A/B too (see #34's comments), so a single local run here isn't worth reporting as a result. That's what this branch's actual CI run is for — same as #34, this PR is the A/B infrastructure, not a claimed win.

### DIFF
--- a/.github/workflows/bsd-kqueue-benchmarks.yml
+++ b/.github/workflows/bsd-kqueue-benchmarks.yml
@@ -313,6 +313,15 @@ jobs:
           run_bench "CWIST (classic)" "env CWIST_C1M_MODE=0 /tmp/cwist_server" 9091 "/tmp/cwist.txt" "/tmp/cwist_stat.txt" 10
           run_bench "CWIST C1M" "env CWIST_C1M_MODE=1 /tmp/cwist_server" 9091 "/tmp/cwist_c1m.txt" "/tmp/cwist_c1m_stat.txt" 10
 
+          # Experimental: cap glibc's per-process malloc arena count to 1
+          # before the fork loop (src/sys/app/app.c), inherited by every
+          # worker (see issue #25's mimalloc writeup - this targets the same
+          # "N processes x M arenas" RSS/tail-latency multiplication mimalloc
+          # tried to fix, without mimalloc's eager per-process reservation).
+          # Same binary/config as "CWIST C1M" above; the env var is the only
+          # variable, so this is a direct A/B, not a new configuration.
+          run_bench "CWIST C1M (arena_max=1)" "env CWIST_C1M_MODE=1 CWIST_MALLOC_ARENA_MAX=1 /tmp/cwist_server" 9091 "/tmp/cwist_c1m_arena1.txt" "/tmp/cwist_c1m_arena1_stat.txt" 10
+
           # Tuned low-latency profile (CWIST classic only): lower connection
           # pressure (c100, 4 wrk threads) so the per-request latency floor is
           # visible alongside the c400 throughput numbers.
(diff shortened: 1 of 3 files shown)


gg582:
## A/B/C: glibc default vs. mimalloc vs. `CWIST_MALLOC_ARENA_MAX=1`

Both allocator experiments (#34, #35) target the same underlying mechanism — glibc's own C1M-mode arena count exploding as N worker processes each independently accumulate M arenas from internal thread contention. Combining their CI numbers here for a side-by-side, since they answer the same question two different ways.

**Caveat**: these are from two separate CI runs (#34's mimalloc run and this PR's arena_max run), not one run with all three measured back to back. Both landed on the same runner class (4 vCPU AMD EPYC 7763) and the "glibc default" baseline came out close enough between the two runs (RPS 117,431 vs. 117,229; P99.999 24.18ms vs. 28.58ms — within the kind of run-to-run spread this shared-runner suite already documents) that comparing across them is reasonable, but not as clean as a true single-run A/B/C would be.

| Metric | A: glibc default | B: mimalloc (#34) | C: arena_max=1 (this PR) |
|---|---|---|---|
| RPS | 117,229–117,431 | 115,279 | **119,245** |
| P90 | 6.33–6.65ms | 6.88ms | **6.29ms** |
| P99 | 14.56–15.47ms | 16.27ms | **13.63ms** |
| P99.999 | 24.18–28.58ms | 47.61ms | **25.26ms** |
| RSS | 10.25–10.32 MB | 22.7 MB | **7.42 MB** |

**C wins on every metric, against both A and B.** B (mimalloc) is worse than baseline A on every metric it was measured against. C isn't just "better than the failed experiment" — it beats the unmodified default too, on RPS, P99, P99.999, and RSS, with P90 roughly a wash.

Same mechanism, opposite outcomes, because of *how* each one addresses it: mimalloc pays an eager per-process arena reservation to solve a problem that doesn't need reserving anything, while `mallopt(M_ARENA_MAX, 1)` just stops glibc's own arena count from multiplying in the first place — zero new dependency, zero upfront cost, inherited by every forked worker for free via `fork()`'s normal process-state copy.

Given C wins outright rather than trading one metric for another, this reads less like "pick a tuning knob" and more like "the default should probably just be this" — worth deciding once this PR's own single-run numbers are corroborated by a rerun, rather than treating a single CI pass as final. Flagging that instead of proposing it outright here.

PR #39, opened 2026-09-12
perf(http): optional classic-pool lock sharding (CWIST_POOL_SHARDS)

gg582:
## Context

Follow-up to #38 (classic-pool starvation fix). While reviewing that PR, a
question came up: could per-worker OS-scheduling variance still cause
unfairness under the shared queue, even with #38's `pending > idle` fix?

Tracing `http_dynamic_worker_thread`/`cwist_http_pool_submit` answers that
directly: there is no per-worker dispatch decision to be unfair about. All
idle workers block on one shared `pthread_cond_t` and race to pop the head
of one shared FIFO; #38's fix is an aggregate (queue depth vs. idle count)
decision, independent of which specific worker gets woken. So that
fairness concern is already closed by #38 - no code change needed for it.

That left a different, real question: is the single global mutex
protecting the shared queue itself a submitter-side bottleneck at high
fan-out? This PR is the experiment answering that question. It is **not**
a fix for anything in #38, and **not** expected to move the-benchmarker's
P99.999 number (same caveat as #37 - different bottleneck than what a
trivial `GET /` handler exercises).

## What changed

`CWIST_POOL_SHARDS=N` (default 1 = current behavior, byte-identical hot
path) splits the classic pool's queue/lock/cond/counters into N
independent shards; submitters round-robin across them. Shard 0 is always
the existing static `g_dyn_pool`, so the default path is unchanged.

Full design, real local measurements (~2.4x submit throughput at N=12 on
12 cores / 32 contending submitters), and honest limitations (local-only,
not CI-validated, not an HTTP-level result yet) are in
`docs/classic-pool-sharding.md`.

## A real bug found and fixed along the way

ASan caught a genuine heap-use-after-free in the first version of this
change: `cwist_http_pool_destroy()` freed a shard's backing memory right
after broadcasting shutdown, without waiting for its detached worker
threads to actually wake and return. Fixed by having each sharded worker
(shard 0 excluded - it's static and never freed, preserving
`test_classic_pool_scaling.c`'s existing `active_workers == thread_count`
assertion) signal completion via `active_workers`, and having `destroy()`
wait (bounded to 5s) on that per extra shard before freeing anything.
Documented in the same doc for the record.

## Testing

- New `tests/test_pool_sharding.c` (added to `make test`): functional
  correctness across shard counts 1/4/8/64, including edge cases (shard
  count not a divisor of thread count, shard count exceeding worker
  count). Every submitted task must run exactly once.
- `make test` (full suite) passes clean, no regressions.
- Both `test_pool_sharding` and the pre-existing `test_classic_pool_scaling`
  pass clean under `SANITIZE=address,undefined`.
- `tests/bench_pool_contention.c` (manual only, not wired into `make test`)
  is the throughput probe behind the numbers in the doc.

## Status

Experimental / opt-in / default off. Open questions (real CI benchmark run,
a sensible default N if ever promoted, cross-shard FIFO ordering trade-off)
are listed at the end of `docs/classic-pool-sharding.md`.

### DIFF
--- a/.github/workflows/bsd-kqueue-benchmarks.yml
+++ b/.github/workflows/bsd-kqueue-benchmarks.yml
@@ -311,6 +311,18 @@ jobs:
           # CWIST is measured twice: the classic thread-pool path (the
           # ultra-low-latency profile) and the event-driven C1M reactor path.
           run_bench "CWIST (classic)" "env CWIST_C1M_MODE=0 /tmp/cwist_server" 9091 "/tmp/cwist.txt" "/tmp/cwist_stat.txt" 10
+
+          # Experimental: split the classic pool's single shared queue/lock
+          # into N independent shards (src/net/http/http.c), submitters
+          # round-robin across them. Answers a different question than the
+          # arena_max/mimalloc experiments below - not an allocator change,
+          # a submitter-side lock-contention one (see
+          # docs/classic-pool-sharding.md and issue #25). Same binary/config
+          # as "CWIST (classic)" above, CWIST_POOL_SHARDS is the only
+          # variable, so this is a direct A/B against that row, not a new
+          # configuration. N = nproc on this runner.
+          run_bench "CWIST (classic, sharded pool)" "env CWIST_C1M_MODE=0 CWIST_POOL_SHARDS=$(nproc) /tmp/cwist_server" 9091 "/tmp/cwist_sharded.txt" "/tmp/cwist_sharded_stat.txt" 10
+
           run_bench "CWIST C1M" "env CWIST_C1M_MODE=1 /tmp/cwist_server" 9091 "/tmp/cwist_c1m.txt" "/tmp/cwist_c1m_stat.txt" 10
 
           # Experimental: cap glibc's per-process malloc arena count to 1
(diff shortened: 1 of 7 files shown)


gg582:
## Classic-pool lock sharding (`CWIST_POOL_SHARDS`) - disproven at HTTP level

Follow-up on #38's starvation fix raised a fairness question (does OS
scheduling variance between workers still cause unfairness under the
shared queue?). Tracing the code closed that question directly: the
classic pool has no per-worker dispatch to be unfair about - all idle
workers race one shared `pthread_cond_t`/FIFO, and #38's `pending > idle`
check is an aggregate decision independent of which worker gets woken.

That left a different question worth testing: is the single global mutex
on that shared queue a submitter-side bottleneck at high fan-out? PR #39
(`experiment/25-pool-lock-sharding`) tested this.

**Local microbenchmark** (lock-only, synthetic, 12 cores): up to ~2.4x
submit throughput at `CWIST_POOL_SHARDS=12` under 32 contending submitter
threads.

**Real CI A/B** (`wrk -t12 -c400 -d10s`, 4 vCPU runner, `N=4` shards) tells
the opposite story:

| metric | classic (baseline) | classic (sharded, N=4) | change |
|---|---:|---:|---:|
| RPS | 121,575 | 123,919 | +1.9% (noise) |
| P99 | 6.85ms | 8.10ms | **+18.2% worse** |
| P99.999 | 25.46ms | 30.01ms | **+17.9% worse** |
| RSS | 17,024 KiB | 10,556 KiB | -38% (unexplained) |

Real request handling (accept/parse/write) dominates the critical
section's cost enough that removing lock contention barely moves
throughput, while splitting one pooled queue into N smaller ones costs
tail latency - a shard that gets a bursty share of the round-robin queues
up behind its own backlog while another shard's worker sits idle, a
pooling-effect cost the single-queue design didn't have. Same
negative-result shape as the mimalloc experiment earlier in this issue.

**Conclusion: `CWIST_POOL_SHARDS` stays in the tree as a documented,
opt-in, default-off option (correctness is real, covered by
`tests/test_pool_sharding.c`), but is not promoted and not recommended.**
Full writeup: `docs/classic-pool-sharding.md`.

PR #41, opened 2026-09-12
fix(sstring): synchronize string size and detach borrowed buffers in rtrim and ltrim

thegoodengineer:
## Summary
This PR fixes two related correctness and memory safety issues in `cwist_sstring` trimming:

1. **Size desynchronization in `rtrim`**:
   - `cwist_sstring_rtrim()` set `str->data[end + 1] = '\0'` when trimming trailing whitespace, but never updated `str->size`.
   - As a result, `cwist_sstring_get_size()` and any subsequent operations relying on `str->size` (e.g. `append`, `change_size`, copy) remained stale with the pre-trimmed length.
   - Fixed by updating `str->size = end + 1` after truncation.

2. **Borrowed buffer detachment**:
   - When `str->borrows_buffer` was true (such as from `cwist_sstring_borrow()`), `cwist_sstring_rtrim()` and `cwist_sstring_ltrim()` previously mutated `str->data` in place. This violated the immutability guarantee for external/static storage and would trigger a crash if the borrowed buffer resided in read-only memory.
   - Fixed by checking `str->borrows_buffer` and cleanly detaching the buffer via `cwist_sstring_reserve()` before mutating.

3. **Tests**:
   - Extended `test_trim()` in `tests/test_sstring.c` to verify `cwist_sstring_get_size()` matches the post-trimmed length for both `trim` and `rtrim`.
   - Added verification for whitespace-only strings (`"    "` -> empty string, size 0).
   - Added verification that trimming borrowed buffers safely detaches without mutating the caller's original borrowed memory.

## Testing
- Unit tests compiled and passed without warnings or errors. Target branch: `dev`.

### DIFF
--- a/src/core/sstring/sstring.c
+++ b/src/core/sstring/sstring.c
@@ -267,8 +267,19 @@ cwist_error_t cwist_sstring_ltrim(cwist_sstring *str) {
     }
 
     if (start > 0) {
-        memmove(str->data, str->data + start, len - start + 1);
-        str->size -= start; 
+        if (str->borrows_buffer) {
+            char *new_data = cwist_sstring_reserve(str, len - start, 0);
+            if (!new_data) {
+                err.error.err_i8 = ERR_SSTRING_NULL_STRING;
+                return err;
+            }
+            memcpy(new_data, str->data + start, len - start + 1);
+            str->data = new_data;
+            str->borrows_buffer = false;
+        } else {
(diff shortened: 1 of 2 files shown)


gg582:
woah, I even couldn't imagine if there are unresolved bugs at sstring...Since it's a 'safe string' it should have been resolved earlier. It's my huge mistake. Thanks for finding this bug.

PR #42, opened 2026-09-12
fix(json_heal): preserve commas inside string literals during trailing comma healing

thegoodengineer:
## Summary
This PR fixes an issue in \`cwist_json_heal\` where trailing comma removal erroneously stripped commas from inside JSON string literals.

### Details
- In \`src/core/utils/json_heal.c\`, \`remove_trailing_commas()\` scans for \`,\` before closing brackets (\`]\`) or braces (\`}\`).
- Previously, it did not track string literal context or escape sequences (\`in_str\` and \`esc\`), unlike \`balance_brackets()\` and the comment stripper in the same pipeline.
- As a result, valid string literals containing commas immediately followed by closing braces/brackets (e.g. \`{\"query\": \"SELECT a, } FROM tbl\"}\` or \`{\"pattern\": \"[a, ]\"}\`) had their internal commas removed, corrupting the payload.
- Fixed by tracking string state (\`in_str\`, \`esc\`) inside \`remove_trailing_commas()\`, ignoring characters inside quotes.

### Tests
- Added \`test_l1_string_with_comma_and_closer()\` to \`tests/test_json_heal.c\` validating that string literals with commas followed by braces/brackets remain intact while true trailing commas outside strings are safely cleaned up.

### DIFF
--- a/src/core/utils/json_heal.c
+++ b/src/core/utils/json_heal.c
@@ -79,8 +79,25 @@ static void log_append(char *log, size_t log_sz, const char *msg) {
 static void remove_trailing_commas(strbuf_t *b) {
     char  *d   = b->data;
     size_t len = b->len;
-    for (size_t i = 1; i < len; i++) {
-        if (d[i] == ']' || d[i] == '}') {
+    bool in_str = false;
+    bool esc = false;
+    for (size_t i = 0; i < len; i++) {
+        char c = d[i];
+        if (esc) {
+            esc = false;
+            continue;
+        }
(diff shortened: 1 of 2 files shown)


gg582:
That's much better. Many json data in a real world contains escape sequence and `"` character(and sometimes both are used like `\"`. That was quite common at Professor Jin Sung-geun's Mobile Network Labs.

If json heal fails at a monitoring server using CWIST it can be a huge accident.
Thank you very much.

PR #43, opened 2026-09-12
fix(seq): preserve max_data_len limit across assembler resets

thegoodengineer:
## Summary
This PR preserves the configured \`max_data_len\` memory ceiling across \`cwist_seq_assembler_reset()\`.

### Details
- When a sequenced-message assembler is created via \`cwist_seq_assembler_create_limited(max_data_len)\`, callers specify a maximum reassembled payload limit to protect against memory exhaustion from untrusted peers.
- In \`src/core/seq/seq.c\`, \`cwist_seq_assembler_reset()\` called \`memset(a, 0, sizeof(*a))\`, which inadvertently zeroed out \`a->max_data_len\`.
- Consequently, after a reset, the assembler reverted to the default unconstrained ceiling (\`CWIST_SEQ_MAX_REASSEMBLED_SIZE\`, 64 MiB), bypassing the caller's configured resource limit.
- Fixed by preserving \`a->max_data_len\` across the reset.

### Tests
- Added \`test_reset_preserves_limit()\` in \`tests/test_seq.c\` validating that resetting a limited assembler preserves its \`max_data_len\` limit and continues rejecting chunks from messages that exceed this limit.

### DIFF
--- a/src/core/seq/seq.c
+++ b/src/core/seq/seq.c
@@ -150,10 +150,12 @@ void cwist_seq_assembler_destroy(cwist_seq_assembler_t *a) {
 
 void cwist_seq_assembler_reset(cwist_seq_assembler_t *a) {
     if (!a) return;
+    size_t max_data_len = a->max_data_len;
     cwist_free(a->data);
     cwist_free(a->received);
     cwist_free(a->payload_lens);
     memset(a, 0, sizeof(*a));
+    a->max_data_len = max_data_len;
 }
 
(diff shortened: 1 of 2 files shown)


src/core/seq/seq.c
gg582, line 1: Well, that's really crucial.

Previous behavior resets the data via `memset(a, 0, sizeof(*a));`.

Then `a->max_data_len` goes zero. The data is not preserved. but this patch definitely solves it.

PR #44, opened 2026-09-12
fix(cookie): validate hex digits in cookie decode, fix pointer underflow in parser, and add tests

thegoodengineer:
## Summary
This PR fixes decoding robustness and parser pointer safety in \`cwist_cookie\`, and adds a dedicated unit test suite for the cookie module.

### Details
1. **Hex validation in \`cwist_cookie_decode()\`**:
   - Previously, \`cwist_cookie_decode()\` called \`sscanf(in + i + 1, "%2x", &hex)\` whenever it encountered \`%\` without verifying that both following characters were valid hexadecimal digits.
   - For incomplete/malformed sequences such as \`%1g\`, \`sscanf\` matched the single hex digit \`1\` and consumed 2 characters, skipping \`g\` and corrupting output.
   - Fixed by requiring \`isxdigit()\` on both characters following \`%\`, leaving non-hex sequences as safe literals.

2. **Pointer bounds safety in \`cwist_cookie_parse()\`**:
   - Trimming trailing whitespace on \`name\` computed \`char *end = name + strlen(name) - 1\`. When \`name\` was empty (e.g. \`; =val\`), \`name - 1\` resulted in out-of-bounds pointer arithmetic before the array boundary (undefined behavior in C).
   - Fixed by using safe length-based index bounds checks and ignoring empty cookie names.

3. **New unit test suite and Makefile integration**:
   - Added \`tests/test_cookie.c\` covering:
     - URL-encoding and decoding (valid hex, lowercase/uppercase, \`+\` as space, invalid/incomplete percent sequences, small buffer overflow protection)
     - Cookie header parsing into \`cwist_query_map\` (whitespace trimming, multiple cookies, empty tokens)
     - Cookie setting with full options (\`Path\`, \`Domain\`, \`Max-Age\`, \`HttpOnly\`, \`Secure\`, \`SameSite\`) and deletion (\`Max-Age=0\`).
   - Wired \`test_cookie\` into \`TEST_TARGETS\` in \`Makefile\`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -487,6 +487,7 @@ TEST_TARGETS = test_classic_pool_scaling \
                test_static_and_range \
                test_session \
                test_csrf \
+               test_cookie \
                test_waf \
                test_db_pool \
                test_db_memory \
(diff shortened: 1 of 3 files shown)


gg582:
That's a good point. I guess I need to dig more for decoding robustness.

gg582:
LGTM

PR #48, opened 2026-09-13
fix(app): keep the hard file limit at startup

DPS0340:
## Summary

Related to #25. This is a partial C1M fix.

- Keep the existing hard file limit.
- Increase only the soft limit, within the hard limit.
- Keep a higher soft limit. Report failed system calls without a success message.

The old code overwrote both limits with 1050000. Its fallback then reused the overwritten value. With Docker limits of 1024/524288, both calls failed and the soft limit stayed at 1024.

## Checks

- Real HTTP control, one process and one worker, 40 open connections: base had 32 HTTP 200 and 8 HTTP 503; this change had 40 HTTP 200.
- Three real child-process cases and eight system-call cases pass.
- Normal and ASan/UBSan resource checks pass with and without NDEBUG. The original-code NDEBUG control fails at the expected runtime check.
- Normal Linux ARM64 suite: 58/59 Make targets. The same gRPC append assertion fails on base d08caace.
- Targeted HTTP, pipeline, and deferred-work sanitizer checks pass.
- Independent review approved the four committed files.

The P99.999 comparison is still in progress. This PR does not claim a tail-latency gain or close #25.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -453,7 +453,8 @@ $(CNATS_LIB):
 
 # --- Test Targets ---
 
-TEST_TARGETS = test_classic_pool_scaling \
+TEST_TARGETS = test_app_resource_limits \
+               test_classic_pool_scaling \
                test_reactor_wake \
                test_sstring \
                test_seq \
@@ -529,6 +530,10 @@ bench_security_pool: $(LIB_NAME) tests/bench_security_pool.c
 	./bench_security_pool
 
 test: $(TEST_TARGETS)
+
+test_app_resource_limits: $(LIB_NAME) tests/test_app_resource_limits.c
+	$(CC) $(CFLAGS) -o $@ tests/test_app_resource_limits.c $(LIB_NAME) $(LIBS)
+	./$@
 
 test_classic_pool_scaling: $(LIB_NAME) tests/test_classic_pool_scaling.c
 	$(CC) $(CFLAGS) -o $@ tests/test_classic_pool_scaling.c $(LIB_NAME) $(LIBS)
--- a/docs/c1m-file-limits.md
+++ b/docs/c1m-file-limits.md
@@ -21,6 +21,25 @@ A failed system call produces an error message, not a success message.
 With this change, the test container used a soft limit of 524288.
 All 40 connections received HTTP 200. The hard limit did not change.
 
+## Tuning
+
+The 1050000 target is a compile-time default (`CWIST_DEFAULT_FD_LIMIT_TARGET`
+in `src/sys/app/app.c`), overridable via `CWIST_FD_LIMIT_TARGET` - matching
+the project's convention for this kind of knob (`CWIST_MALLOC_ARENA_MAX`,
+`CWIST_POOL_SHARDS`, `CWIST_HTTP_YIELD_BATCH`, ...). Set it when the default
+doesn't fit a deployment: a smaller container quota, or a workload that opens
+more than one fd per connection (proxying, per-connection temp/log files).
+
+Any value that isn't a valid positive integer (unset, empty, non-numeric,
+trailing garbage, zero, negative) falls back to the default - `cwist_app_tune_system()`
+never runs with `target == 0` or a partially-parsed number. The hard-limit
+clamp and no-op-if-already-sufficient behavior above apply identically
+whether the target came from the env var or the default.
+
+```
+CWIST_FD_LIMIT_TARGET=200000 ./your_cwist_app
+```
+
 ## Checks
 
 - Three child-process checks use real file limits.
--- a/docs/c1m-file-limits.md
+++ b/docs/c1m-file-limits.md
@@ -1,0 +1,34 @@
+# C1M file limits
+
+This is a partial fix for issue #25.
+
+## Cause
+
+The old code sets both file limits to 1050000 before it calls `setrlimit`.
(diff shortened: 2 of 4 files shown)


gg582:
Looks good to me, I guess we can pull out 105000 magic number as a tunable env variable.

gg582:
Pushed a follow-up commit making the fd-limit target tunable:

`CWIST_FD_LIMIT_TARGET` env var, defaulting to the existing 1050000
(`CWIST_DEFAULT_FD_LIMIT_TARGET`) when unset - zero behavior change for
anyone not setting it. Matches the project's existing convention for this
kind of knob (`CWIST_MALLOC_ARENA_MAX`, `CWIST_POOL_SHARDS`,
`CWIST_HTTP_YIELD_BATCH`).

Any value that isn't a valid positive integer (unset, empty, non-numeric,
trailing garbage, zero, negative) falls back to the default rather than
running `cwist_app_tune_system()` with a bad target - covered by 9 new
`check_mock_env` cases.

Verified locally: full `test_app_resource_limits` passes both plain and
under `SANITIZE=address,undefined` (explicitly forced a rebuild of `app.o`
under sanitizers - `make SANITIZE=... libcwist.a` alone won't force-rebuild
an already-current object file, easy to accidentally test the wrong
binary).

DPS0340:
## Follow-up: completed 81-run local comparison (partial fix only)

The FD-limit regression is supported by the full comparison, but this is **not a claim that #25 is solved**.

- 3 implementations (baseline, FD fix, Axum) × 3 routes × 3 connection counts × 3 repetitions = **81 runs**. Implementation order rotated between repetitions. All 81 histograms decoded and their counts matched the reported request counts.
- Baseline at c512: all 9 runs contained errors, **476 non-success responses total**. These runs are excluded from successful-response latency comparisons.
- FD-fixed CWIST: **27/27 error-free runs**. Axum: **27/27 error-free runs**. Baseline c64/c256: 18/18 error-free runs.
- Median of the nine individual run P99.999 values at each concurrency (three routes × three repetitions; **not a pooled percentile**):
  - c64: baseline 0.957 ms; FD-fixed 1.169 ms; Axum 1.984 ms.
  - c256: baseline 6.979 ms; FD-fixed 6.211 ms; Axum 6.911 ms.
  - c512: baseline excluded due to errors; FD-fixed 6.683 ms; Axum 8.199 ms.

There is no consistent P99.999 improvement attributable to the FD fix on the already-error-free baseline cases. The supported outcome is removal of admission-limit failures.

### Conditions and revision limits

Dedicated ARM64 Linux VM, 4 CPUs / 4 GiB; pinned Ubuntu 24.04 Docker image; ZRK 2.5.0; Docker bridge; actual CWIST backend verified as **epoll**. Each run used `-c 64/256/512 -d 15s -R1000:100000`, with 60s startup and one 5s / c50 warmup per implementation group. Warmup used the later benchmark revision's default rate, not the original result's ramped warmup. Historical hardware and executed binaries are not reproduced.

Baseline source: `d08caacebede1f05692e4f8e5f1283e1dc5deb0f`. The measured candidate's production delta is the FD clamp in original PR head `8cafc397de0de051ecf08837c87fabc97f5e7c78`. The later `CWIST_FD_LIMIT_TARGET` follow-up (`8a724e7b9ab44dd347922a541dbd37938f418362`) was **not** included in these local measurements; its separate CI is green.

### Benchmark-label correction

The original public `percentile99999` field imports **P99.99**, then averages route-level scalar values. It is not true P99.999. This comparison instead decodes P99.999 from each emitted HDR histogram. Sources:
- [collector mapping](https://github.com/the-benchmarker/web-frameworks/blob/ba951edd2641fbe751602e7c92ed737e5d1cc643/.tasks/collect.rake#L73)
- [SQL aggregation](https://github.com/the-benchmarker/web-frameworks/blob/ba951edd2641fbe751602e7c92ed737e5d1cc643/.tasks/db.rake#L13)
- [original warmup](https://github.com/the-benchmarker/web-frameworks/blob/676d84f12392827e2ef364837b129b5cd4af02ed/.tasks/config.rake#L139)

The original PR head's separate CI `wrk -t12 -c400 -d10s` artifact still reports C1M P99.999 **51.552 ms** vs Axum **18.331 ms** ([run](https://github.com/c4punks/CWIST/actions/runs/34744538671)). That is a different, saturated workload on a shared x86 runner; it must not be pooled with this rate-limited ARM64 comparison. The remaining investigation needs a controlled backend/topology comparison on that workload rather than treating the successful FD repair as a tail-latency solution.

Post-run isolation caveat: dormant prefork `stress_test` children from an earlier test run were found in the VM (`io_cqring_wait`) and have now been terminated using identity-checked pidfds. These descriptive latency numbers are not acceptance evidence from a perfectly idle VM. A clean backend/topology experiment is still required. The deterministic held-connection FD regression remains a separate correctness result.

gg582:
Anyway, following CWIST's convention, changing hard file limit was dangerous. Also, hard-coded magic number is removed. This pull request is not changing a whole structure(so it's a basic refactoring task). IMO it's ready to merge.

gg582:
Left a correction on #25 acknowledging that several of my earlier comments there cited the-benchmarker's `percentile99999` field as P99.999 when it's actually P99.99 (per the finding above) - see https://github.com/c4punks/CWIST/issues/25#issuecomment-5652089823

PR #49, opened 2026-09-13
fix(jwt): reject base64url remainder of 1 and validate chunk feed

thegoodengineer:
### Summary of Changes

1. **Base64URL remainder validation (`RFC 4648 §4`)**:
   - In `b64url_decode()`, a Base64 remainder of 1 6-bit character (`padded_len - i == 1`) is mathematically impossible for valid Base64 data (a single 6-bit unit cannot encode an 8-bit byte).
   - Previously, if `rem == 1`, the function fell through the `if (rem == 2)` / `else if (rem == 3)` blocks and returned the partially decoded buffer as valid.
   - Now explicitly checks `else if (rem == 1)` and frees the buffer and returns `NULL`.

2. **Chunk feed error handling in `cwist_jwt_join_chunks()`**:
   - `cwist_seq_assembler_feed()` returns `false` if chunk sequence validation fails or the chunk cannot be accepted.
   - Previously, the return value was ignored and parsing continued.
   - Now checks the return value; if feeding fails, cleans up the assembler and returns `NULL`.

3. **Unit Tests**:
   - Added `test_malformed_b64url_remainder()` to `tests/test_jwt.c` to verify that tokens with remainder 1 are rejected.
   - Added `test_corrupt_chunk_feed()` to `tests/test_jwt.c` to verify that chunk join properly fails when feeding corrupted chunks.

### DIFF
--- a/src/security/jwt/jwt.c
+++ b/src/security/jwt/jwt.c
@@ -118,6 +118,11 @@ static unsigned char *b64url_decode(const char *src, size_t src_len, size_t *out
         if (a < 0 || b < 0 || c < 0) { cwist_free(out); return NULL; }
         out[j++] = (unsigned char)((a << 2) | (b >> 4));
         out[j++] = (unsigned char)((b << 4) | (c >> 2));
+    } else if (rem == 1) {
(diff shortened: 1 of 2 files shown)


gg582:
LGTM :+1:

PR #50, opened 2026-09-13
fix(test_client): use reentrant strtok_r and add null checks for cookie parsing

thegoodengineer:
## Summary
This PR improves thread safety and null robustness in the test client's cookie jar implementation:

1. **Reentrant cookie header parsing**:
   - `cookie_jar_update()` previously used standard `strtok(buf, ";")` and `strtok(NULL, ";")`.
   - `strtok` uses a static internal buffer that is not thread-safe and can cause race conditions or collide with other concurrent string tokenization in test harnesses.
   - Replaced with POSIX reentrant `strtok_r()` using a local stack `saveptr`.

2. **Null safety in cookie handling**:
   - Guarded `find_cookie()` against NULL `name` and NULL `head->name`.
   - Checked allocation returns for `clone_str(trim(name_val))` and `clone_str(trim(eq + 1))` in `cookie_jar_update()`.
   - In `cookie_jar_apply()`, validated that `c->name`, `c->value`, `adhoc[i].key`, and `adhoc[i].value` are non-null before computing buffer length and writing into the `Cookie` header string.
   - Avoided empty or malformed `Cookie` header emission if all cookies or adhoc entries are NULL.

3. **Tests**:
   - Added test cases in `tests/test_test_client.c` for `cwist_test_client_get_cookie(client, NULL)`.
   - Added test cases for `adhoc` cookies containing NULL key or NULL value, asserting graceful filtering without crashing.

## Testing
- Verified on Linux/WSL with gcc. All test client assertions passed. Target branch: `dev`.

### DIFF
--- a/src/sys/app/test_client.c
+++ b/src/sys/app/test_client.c
@@ -27,24 +27,35 @@ static void free_cookie(cwist_test_client_cookie *c) {
 }
 
 static cwist_test_client_cookie *find_cookie(cwist_test_client_cookie *head, const char *name) {
+    if (!name) return NULL;
     for (; head; head = head->next) {
-        if (strcmp(head->name, name) == 0) return head;
+        if (head->name && strcmp(head->name, name) == 0) return head;
     }
     return NULL;
 }
 
 static void cookie_jar_apply(cwist_test_client *client, cwist_http_request *req,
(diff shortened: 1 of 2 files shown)


gg582:
Classic null checks, `strtok`-> `strtok_r`..
Those defensive blocks are important for web frameworks :)

thanks

PR #51, opened 2026-09-13
fix(sstring): correct html escape entity typo for > and avoid redundant allocations

thegoodengineer:
## Summary
This PR fixes an HTML escaping corruption bug and avoids redundant heap allocations in `cwist_sstring_append_escaped`:

1. **Fix entity typo for `>` character**:
   - In `src/core/sstring/sstring.c`, escaping `>` wrote `&gtl` instead of the valid HTML entity `&gt;`:
     ```c
     case '>':
         memcpy(ptr, "&gtl", 4); // replaced with "&gt;"
     ```
   - This caused corrupt HTML output whenever greater-than characters were escaped.

2. **Eliminate unnecessary heap allocation on success**:
   - `cwist_sstring_append_escaped()` was unconditionally allocating `err.error.err_json = cJSON_CreateObject()` at the start of the function and freeing it at the end on success.
   - Now, `cJSON_CreateObject()` is only allocated when an actual error occurs (e.g. fixed buffer overflow or out of memory).

3. **Tests**:
   - Added `test_html_escape()` in `tests/test_sstring.c` verifying escaping of `<`, `>`, `&`, `"`, `'`, standalone `>`, and handling of NULL pointers.

## Testing
- Successfully compiled and ran `tests/test_sstring.c` on Linux/WSL. All tests passed. Target branch: `dev`.

### DIFF
--- a/src/core/sstring/sstring.c
+++ b/src/core/sstring/sstring.c
@@ -521,19 +521,24 @@ cwist_error_t cwist_sstring_append_escaped(cwist_sstring *str, const char *data)
     size_t input_len = strlen(data);
     size_t new_size = current_len + (input_len * 6) + 1;
 
-    cwist_error_t err = make_error(CWIST_ERR_JSON);
-    err.error.err_json = cJSON_CreateObject();
-
     if (str->is_fixed) {
         if (new_size > str->size) {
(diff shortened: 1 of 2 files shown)


gg582:
That's a typo! I didn't know it even I am running my dev blog server at https://oborona.zip

Anyway that's really valid. I don't sure why didn't it make a problem earlier.

Thanks

PR #52, opened 2026-09-13
fix(html): escape attribute values and support boolean and numeric attributes safely

thegoodengineer:
## Summary
This PR fixes XSS / attribute injection vulnerabilities and adds safe handling of boolean, numeric, and null attributes in `cwist_html_builder`:

1. **HTML attribute value escaping**:
   - `render_element()` previously emitted attribute values directly via `cwist_sstring_append(out, attr->valuestring)`.
   - Any attribute containing `"` (double quotes), `<`, `>`, or `&` would break out of the HTML attribute quotes or produce invalid/vulnerable markup.
   - Now, attribute string values are safely escaped with `cwist_sstring_append_escaped()`.

2. **Safe handling of non-string and boolean attributes**:
   - In HTML, attributes can be boolean flags (e.g. `disabled`, `checked`, `readonly`) or numeric (e.g. `tabindex="2"`).
   - If a non-string cJSON item was added to `el->attributes`, `attr->valuestring` was NULL, leading to empty/malformed rendering.
   - Handled `cJSON_IsTrue(attr)` to emit standard boolean HTML attributes (` disabled`).
   - Handled `cJSON_IsNumber(attr)` to serialize double/integer values (` tabindex="2"`).

3. **Escaping inner text and null safety**:
   - Ensured element `inner_text` is escaped with `cwist_sstring_append_escaped()`.
   - Added null guards in `cwist_html_element_create()`, `cwist_html_element_add_attr()`, `cwist_html_element_add_class()`, `cwist_html_element_set_text()`, and `render_element()`.

4. **Tests & Build integration**:
   - Added `tests/test_html_builder.c` covering elements, class handling, attribute escaping, boolean/numeric attributes, nested children, and null safety.
   - Added `test_html_builder` to `Makefile`.

## Testing
- Successfully compiled and ran `tests/test_html_builder.c` under WSL/Linux. All assertions passed. Target branch: `dev`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -511,7 +511,8 @@ TEST_TARGETS = test_classic_pool_scaling \
                test_full_gc_ownership_handoff \
                test_defer_free \
                test_proto_gen \
-               test_proto_desc
+               test_proto_desc \
+               test_html_builder
 
 .PHONY: all test $(TEST_TARGETS) fuzz_seq install uninstall dist clean rebuild examples clean-examples wasm wasm-smoke clean-wasm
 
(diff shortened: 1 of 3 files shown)


gg582:
The classic injection problem. And it's solved now. Thanks a lot!

Since its main language is C17 when XSS injection is available many senior developers will say that C is not enough to write a website. But, personally I want to prove that language is not a huge matter.

I am glad that you are considering CWIST as a real web framework. Some people think that it's a joke.

PR #53, opened 2026-09-13
feat(css_composer): support 3-digit hex colors and add strict validation

thegoodengineer:
## Summary
This PR adds support for standard CSS 3-digit hex color shorthands (`#RGB`) and improves parsing robustness in `cwist_color_hex_to_rgb()`:

1. **3-digit `#RGB` hex color support**:
   - In standard CSS (CSS Color Module Level 3 & 4), `#RGB` is shorthand for `#RRGGBB` where each digit is repeated (e.g. `#fff` -> `#ffffff`, `#f00` -> `#ff0000`, `#123` -> `#112233`).
   - Previously, `cwist_color_hex_to_rgb()` only accepted 6-character strings (`strlen(hex) == 6`), causing any 3-digit hex shorthand to fall back to black `{0, 0, 0}`.
   - Now supports both 3-digit and 6-digit hex formats, with or without leading `#`.

2. **Strict hex character validation**:
   - Replaced `sscanf()` with a direct, fast hex digit parser (`hex_char_to_val()`) verifying that all characters are valid hex digits `[0-9a-fA-F]`.
   - Any invalid character safely falls back to default black `{0, 0, 0}` without partial parsing or undefined behavior.

3. **Tests & Build integration**:
   - Added `tests/test_css_composer.c` verifying 6-digit hex parsing, 3-digit hex shorthand expansion, invalid hex strings (wrong length, non-hex chars, NULL), roundtrip formatting with `cwist_color_rgb_to_hex()`, and stylesheet generation.
   - Added `test_css_composer` target to `Makefile`.

## Testing
- Verified on Linux/WSL via gcc. All CSS composer test assertions passed. Target branch: `dev`.


gg582:
Cool, I sometimes use `#fff` when I develop with other frameworks

PR #61, opened 2026-09-13
feat(gc): header-scoped malloc interception (CWIST_INTERCEPT_MALLOC)

gg582:
## Summary

Implements the last open item of v3.5's GC milestone (see the prior
docs-only PR that settled the design direction): `CWIST_INTERCEPT_MALLOC`,
an opt-in, header-scoped redirection of `malloc`/`calloc`/`realloc`/`free`
onto the same full-GC epoch pipeline `cwist_alloc()`/`cwist_free()` already
use.

## Why header-scoped, not `-Wl,--wrap=malloc`

A link-level wrap intercepts *every* `malloc` reference in the final
binary, including vendored dependencies (BoringSSL, lsquic, cnats,
sqlite3) that never include CWIST headers and have no reason to expect a
non-libc allocator underneath them. Concretely: BoringSSL calls
`OPENSSL_cleanse()` to zero secret key material on free, assuming plain
heap semantics — routing that through an epoch-deferred GC arena would be
a security regression, not a performance footnote.

`#define CWIST_INTERCEPT_MALLOC` before including
`<cwist/core/mem/intercept.h>` only ever takes effect in a translation
unit that opts in; vendored dependencies structurally never do (they don't
include CWIST headers at all), excluding them by construction.

## What changed

- `include/cwist/core/mem/intercept.h` (new): shim declarations +
  conditional macro block. Pulls in `<stdlib.h>` before defining the
  macros — a later `#include <stdlib.h>` (this TU's own, or transitive)
  would otherwise have its own `malloc`/`calloc`/`realloc`/`free`
  declarations mangled by the just-activated macros, which reliably fails
  to compile. Tripping stdlib.h's own include guard on any later
  inclusion avoids it (found this the hard way while building the
  benchmark below).
- `src/core/mem/alloc.c`: the four shim implementations. `malloc`/`calloc`
  keep real semantics (uninitialized vs. zeroed — neither becomes
  `cwist_alloc`'s always-zeroed behavior); all four gate on
  `cwist_full_gc_enabled()` (the same relaxed atomic check
  `cwist_alloc()`/`cwist_free()` already pay) and reuse the existing
  `cwist_gc_scope_track`/`untrack` pipeline rather than a new policy.
  `realloc`'s tracking follows the incoming pointer (untrack old, track
  new, only if the old one was tracked) rather than unconditionally
  opting every result in.
- `tests/test_malloc_intercept.c`: passthrough correctness when full-GC is
  off, tracked-and-swept-without-an-explicit-`free()` when it's on
  (`cwist_gc_scope_flush()`, no second thread needed to observe it), and
  realloc tracking correctly following a disowned pointer.
- `tests/bench_malloc_intercept.c` (manual, not wired into `make test`):
  the actual question this PR needs answered - is "zero cost when off"
  true, and what's the real cost when on?

## Measured overhead

Single-threaded, mixed 16–512 byte allocations with an occasional
`realloc` (see the bench file for exact methodology), this machine:

| configuration | ns/op | vs. baseline |
|---|---:|---:|
| baseline (no shim) | ~11.8 | — |
| shim, full-GC off (default) | ~12.2 | **+3%** |
| shim, full-GC on | ~21.9 | **+86%** |

The "off" cost is one relaxed atomic load plus one extra function call per
operation — matches `cwist_alloc()`/`cwist_free()`'s existing overhead, so
the "zero(-ish) cost when disabled" principle holds. The "on" cost is the
real, expected price of the pending-sweep-list safety net; ~22ns per
malloc/free pair is still fast in absolute terms for a request-handling
budget, but it's a real ~2x, not free.

## Verification

- `make libcwist.a` builds clean.
- `test_malloc_intercept` (3 checks: GC-off passthrough, GC-on
  tracked-and-swept, realloc-follows-pointer) passes normally and under
  `SANITIZE=address,undefined` (rebuilt `alloc.o` explicitly under
  sanitizers to make sure the right binary was actually tested).
- Full existing full-GC test suite re-run for regressions: `test_defer_free`,
  `test_full_gc_toggle_hardening`, `test_full_gc_sweep`, `test_conn_registry`,
  `test_full_gc_ownership_handoff`, `test_gc_ebr_release`,
  `test_io_queue_full_gc` — all pass, no regressions.

## Docs

`docs/GC.md` section 5 and the `ROADMAP.md` v3.5 entry updated from
"settled design direction, not yet built" to "built", with the measured
numbers and usage example.


gg582:
Filed #65 to track the ~86% per-call malloc/free overhead measured here when full-GC mode is on (see the benchmark table in this PR's description) - it's unprofiled and the fix is out of scope for this PR.

PR #63, opened 2026-09-13
fix(ci): isolate benchmark process lifetimes

DPS0340:
## Summary

Related to #25. This repairs benchmark isolation; it does not claim a runtime P99.999 improvement.

The existing `run_bench` ends each profile with `kill -9 $PID`. CWIST forks worker processes. Killing only their parent leaves workers listening on the reused port. A later C1M profile can therefore share its port with surviving classic workers.

### Reproduced with the real CWIST server

With four workers:

- Start classic: four listening processes.
- Apply the current parent-only SIGKILL: three workers remain, and HTTP requests still return 200.
- Start C1M on the same port: three old classic workers plus four new C1M workers are listening.
- The reproduction cleaned up every owned process afterwards.

This proves the isolation defect. It does not reconstruct how many requests in a historical CI run reached each old worker.

## Change

- Start each server and workload in separate private process groups.
- Keep both leaders unreaped until all group signals finish, preventing numeric PGID reuse during cleanup.
- Clean both groups on completion, command failure, timeout, SIGINT, and SIGTERM.
- Bound TERM/KILL cleanup and reject remaining live group members.
- Apply the same supervision to normal and tuned profiles.
- Retain existing workload arguments, warmup, result filenames, and percentile reporting.

The helper is Linux-only, matching the web-server job. It supervises trusted benchmark commands, not programs that deliberately escape their session. Uncatchable termination of the supervisor remains outside its cleanup guarantees.

## Verification

- Eleven Linux integration tests pass normally and with `python3 -O`.
- Cleanup-phase TERM/INT and handler-handoff regressions ensure cancellation cannot be reported as success.
- Covered early server/workload exit with live children, TERM-resistant children, workload failure, timeout, cancellation, and an unrelated process group.
- Real-server GREEN: classic then C1M, four processes each, ten HTTP 200 responses per phase; no live group members or open port after either phase.
- Extracted workflow function smoke test uses real CWIST and a stub load command. It verifies `-t12 -c400 -d10s` and tuned `-t4 -c100 -d10s` argument preservation and cleanup. This smoke test is not a performance benchmark.
- Eleven existing benchmark-renderer tests pass.
- YAML parsing and Bash syntax checks pass.
- Full native-x86 benchmark results are pending this PR's CI. Historical mixed-profile values should not be treated as an isolated baseline for the corrected runs.

## Kept separate

This change does not repair the legacy root-only `ps` statistics or the workload's existing `wrk` error suppression. Neither should be used as proof of error-free or complete scheduling telemetry. No application defaults or latency thresholds change.

### DIFF
--- a/.github/workflows/bsd-kqueue-benchmarks.yml
+++ b/.github/workflows/bsd-kqueue-benchmarks.yml
@@ -24,6 +24,8 @@ jobs:
       - uses: actions/checkout@v4
       - name: Test benchmark report generation
         run: python3 -m unittest discover -s scripts/ci -p 'test_*.py' -v
+      - name: Test benchmark process isolation
+        run: python3 -m unittest discover -s tests -p 'test_benchmark_session.py' -v
 
   bsd-kqueue:
     name: kqueue tests (macOS)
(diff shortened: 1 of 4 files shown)


gg582:
Reviewed the process-group isolation logic and the reproduction in the PR description - the parent-only `kill -9` leaving forked workers alive on the shared port is a real, well-demonstrated bug in the benchmark harness itself, independent of any runtime latency work. Merging on that basis: CI is green (ASan/UBSan, benchmark rendering, HTTP/2 conformance, both benchmark jobs), and this only touches CI/benchmark tooling (workflow YAML + scripts/ci/benchmark_session.py + benchmark_workload.sh), not runtime library code, so the blast radius of being wrong here is a bad benchmark number, not a broken build.

One thing worth flagging for a follow-up rather than blocking this PR on it: since this fixes a real cross-run contamination bug, some of the benchmark numbers already recorded in benchmarks/webserver.json and cited in issue #25 (including comparisons this session made against them) may have been measured on a machine mid-way through starting a subsequent profile while a prior profile's forked workers were still alive on the same port. Worth a note on #25 once this lands, rather than silently treating old and new numbers as comparable.

PR #66, opened 2026-09-13
fix(sstring): prevent overflow in substr and normalize empty string compare

thegoodengineer:
### Summary of Changes
1. **Prevent Signed Integer Overflow in `cwist_sstring_substr()`**:
   - In `src/core/sstring/sstring.c`, `if ((size_t)(start + length) > current_len)` caused signed 32-bit integer overflow (undefined behavior) when `length` was large (e.g. `INT_MAX`).
   - Replaced with `if ((size_t)length > current_len - (size_t)start)` which avoids addition and bounds `length` safely.
2. **Normalize Empty String Comparisons in `cwist_sstring_compare_sstring()`**:
   - A newly created string initialized with `cwist_sstring_create()` has `data == NULL`, while an empty string assigned `""` has `data != NULL` with `data[0] == '\0'`.
   - Comparing these two empty string instances previously returned non-zero (`-1` or `1`), erroneously treating two empty strings as unequal.
   - Normalized both NULL data and empty string data to `""` via `sstring_data_or_empty()` so all empty string representations compare symmetrically and equally (`0`).
3. **Unit Tests**:
   - Added unit test cases to `tests/test_sstring.c` for empty string comparisons (`empty_new` vs `empty_assigned` vs `""` vs `NULL`) and `INT_MAX` length substring extraction.

### Testing
- Tested and verified with GCC under Linux.

### DIFF
--- a/src/core/sstring/sstring.c
+++ b/src/core/sstring/sstring.c
@@ -235,6 +235,10 @@ size_t cwist_sstring_get_size(cwist_sstring *str) {
     return str ? str->size : 0;
 }
 
+static inline const char *sstring_data_or_empty(const cwist_sstring *s) {
+    return (s && s->data) ? s->data : "";
+}
+
 /**
  * @brief Compare two CWIST string objects using strcmp semantics.
  * @param left Left-hand string.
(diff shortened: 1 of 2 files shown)


gg582:
Renamed the new internal helper `sstring_data_or_empty` -> `sstring_as_text` (pushed to this PR's branch). Same static, file-local function, no behavior change - just a clearer name for what it does.

PR #69, opened 2026-09-13
feat(ci): capture identity-bound benchmark diagnostics

DPS0340:
## Scope

Related to #25. This is a diagnostic follow-up stacked on #63, **not a runtime tail-latency fix**. Keep the existing workload flags, warmup, and process-group cleanup unchanged.

The native benchmark currently collects root-only `ps` switch fields that are unsupported on Linux and fall back to zero. Those zeros cannot explain the remaining C1M/Axum tail difference. The startup SQPOLL probe also does not establish each serving reactor's backend.

## Changes

- Capture allowlisted `/proc` observations for the supervised server group immediately before and after the measured workload: PID/TID starttimes, CPU masks, per-thread context switches/runtime, optional scheduler wait, per-process RSS, and anonymous-inode backend counts.
- Produce summaries only for comparable, complete, positive-duration snapshots with stable process/task identities and nonnegative counter deltas. Missing data is `null`; disabled scheduler wait is `null`, not zero.
- Retain legacy output compatibility, explicitly mark legacy context-switch fields unreliable, and save before/after/summary sidecars for all eight profiles. Preserve allowlisted raw artifacts even when a prior step fails.
- Do not read or publish command lines, environments, arbitrary FD targets, or raw exception text. This probe observes a supervisor-owned group; it is not an ownership/security verifier.

## Validation completed locally

- 37 CI-script tests: 23 synthetic probe regressions, 3 functional workload/artifact regressions, and 11 existing report tests.
- 11 existing process-session regressions.
- Both suites pass on Linux normally and with `PYTHONOPTIMIZE=1` / `-O`.
- Workflow YAML and 13 extracted shell blocks pass syntax checks.
- Live supervised Linux C1M run: four processes and twenty stable tasks; each process had five io_uring descriptors and five eventfds, with no eventpoll descriptor. Adjacent snapshots recorded 1,093,270 context switches over approximately 10 seconds; disabled scheduler wait correctly remained null. Server port was closed after cleanup.

The live check used the existing ARM64 test binary, not a new native-x86 performance result. Snapshot overhead lies outside the measured wrk command but may perturb subsequent state. RSS sums double-count shared pages; detected thread turnover or inaccessible data makes summaries incomplete. Threads born and exited entirely between snapshots can be missed, so these are endpoint-observed deltas, not a complete scheduler trace. Context-switch counts alone do not prove the tail's cause. Native workflow results are still pending.

The existing wrk error suppression and Spring AOT training lifecycle are unchanged. A complete proc summary does not establish an error-free workload; inspect raw wrk output separately.

The probe suite was reconstructed after an untracked source file disappeared; current test results refer to the reconstructed staged bytes, not an assertion that the historical test source was recovered identically.


DPS0340:
Native workflow verification for head `90f80af07506272a266ccede13533f2c8554766a` completed successfully: https://github.com/c4punks/CWIST/actions/runs/34751194909

- Retrieved and verified all 24 expected proc sidecars. Raw wrk output for all eight profiles contains completed requests and no reported socket/status-error lines. This is the manually dispatched benchmark workflow, not a claim that every PR check has run.
- Comparable endpoint deltas: C1M, C1M arena1, Axum and Gin (4/8 profiles). Classic had 22 removed and 22 added task identities; tuned classic added two; each Spring profile added one and removed two. Both snapshots themselves were complete. Returning null deltas for these changed task sets is intentional, not a CI failure.
- C1M: 4 processes, 20 tasks, five io_uring descriptors per process, no eventpoll descriptors. Every task within a process shared one singleton CPU mask; the four processes used cores 0/1/2/3. Axum: 1 process, 5 tasks, eventpoll descriptors, mask 0-3.
- Scheduler waiting statistics were disabled, so wait deltas remain null. Observed C1M/Axum context-switch deltas were 280,749/198,440; these endpoint counts alone do not establish a scheduling bottleneck.

In this one 4-vCPU AMD EPYC 7763 run, C1M reported 135,630.62 RPS and wrk-corrected P99.999 30.237 ms; Axum reported 108,818.28 RPS and 19.672 ms. C1M had about 24.6% more throughput and 1.54x the reported tail. The previous run used an Intel Xeon runner, so cross-run changes are not evidence of an optimization. No runtime default was changed and #25 is not resolved.

The next useful experiment is a repeated within-run reactor-thread comparison with worker count, affinity, fixture and load held constant. It has not run yet. Sparse endpoint telemetry is not an atomic scheduler trace or proof of an error-free workload beyond the raw logs inspected here.

PR #70, opened 2026-09-13
fix(reactor): cooperative queuing for foreign-thread posts in a big CQE batch (#25)

gg582:
## Cooperative queuing for foreign-thread posts in a big CQE batch (#25)

Follow-up on the dispatch-round-serialization work already tracked in #25
(yield-batch fix `ce970136`, response coalescing `0bcc4f5a`). Both of
those target **intra-connection** head-of-line blocking. This targets a
different, previously-unaddressed layer: **inter-connection** ordering
within a single io_uring wake round.

### The gap

`cwist_reactor_run()`'s io_uring branch drains a whole CQE batch inline,
synchronously, before calling `reactor_drain_posts()` again. The ring is
4096 entries deep, so a burst of ready connections can put hundreds in
one batch, each served fully before the next even starts.

`reactor_drain_posts()` is what delivers foreign-thread completions
(`cwist_reactor_post()`, used by `cwist_async_defer` for background
job/DB continuations) back to the reactor's owner thread. It was only
ever called once per round, at the *top* of the loop -- i.e. after the
*entire previous* batch had already finished draining. A background job
that completes early in a 512-connection batch waits behind the other
511 connections' full handler execution before its own continuation
runs, even though nothing about that continuation is itself slow.

This reproduces the shape #25 reports: P99.999 scaling sharply with
concurrency while average latency stays flat -- an unlucky arrival time
relative to a big batch, not a slow request.

### The fix

`CWIST_REACTOR_DRAIN_CHUNK=<N>` interleaves `reactor_drain_posts()`
every N connection callbacks instead of only at the batch's end. `0`
(unset, default) is byte-for-byte the previous behavior -- io_uring-
branch-only, no change to the classic pool or epoll paths.

### Measured (before/after, not theorized)

`tests/bench_cooperative_queuing.c`: 512 pipes all made readable before
entering the reactor (so all 512 land as CQEs in one round), each
callback simulating an 80us handler; a background thread posts one
`cwist_reactor_post()` 200us into the round. 10 runs each, this machine:

| configuration | post-to-callback latency |
|---|---:|
| legacy (unset) | 41.1-41.4 ms (~= 512 x 80us, waits for the whole batch) |
| `CWIST_REACTOR_DRAIN_CHUNK=32` | 2.32 ms (~18x) |
| `CWIST_REACTOR_DRAIN_CHUNK=8` | 0.38 ms (~107x) |

`tests/test_reactor_drain_chunk.c` is the pass/fail twin, wired into
`make test`: asserts a chunked post lands in under half the time an
unchunked whole batch would take. Verified clean under ASan+UBSan.
`test_http_fairness` and `test_reactor_wake` both still pass unchanged
(env var unset in their runs).

### What this does NOT fix

Only foreign-thread continuations get this benefit. A plain synchronous
`GET /` benchmark that never calls `cwist_async_defer` has no posts to
interleave, so the-benchmarker-style numbers already in #25 are not
expected to move from this alone. Full writeup, including the two
approaches considered and set aside (time-bounding inline work per
batch; moving handler execution off the reactor thread entirely -- the
"traditional completion queue" the issue thread already cautioned could
regress average latency), is in `docs/cooperative-queuing.md`.

### Tuning

No nonzero default is proposed yet -- this repo's own CI benchmark is a
plain empty-handler workload and can't exercise this path at all. Left
opt-in pending a CI run against an `cwist_async_defer`-using workload.


gg582:
Added a second commit to this PR:

- **arena_max=1** was labeled "Experimental" in the CI job even though PR #35 already adopted it (mimalloc was tried and refuted on real CI data — see issue #25/ROADMAP.md). Reworded the workflow comment, the README generator, and a stale `src/sys/app/app.c` comment that still called it "the next experiment" — it's a decided, adopted fix; this CI leg now exists to keep confirming that decision on every run.
- **drain_chunk=8** (this PR's own fix) is now a direct A/B leg in the real web-server benchmark job, run/parsed/reported exactly like the arena1 leg — not just a local-only micro-benchmark. Scoped honestly in the comment: this workload never calls `cwist_async_defer`, so parity with the plain C1M row is the expected result, not a null finding; the actual tail-latency win stays isolated in `tests/bench_cooperative_queuing.c`.

Merging now.

PR #81, opened 2026-09-13
fix(ci): fix -Warray-compare in test_middleware_jwt and sync malloc-baseline line numbers

thegoodengineer:
### Summary of Changes

1. **Fix `-Werror=array-compare` in `tests/test_middleware_jwt.c`**:
   - In GCC 12+, comparing two arrays directly (`assert(secret1 != secret2);`) triggers `[-Werror=array-compare]`, failing the `Address/UB Sanitizer (build + test)` CI job.
   - Updated to compare array addresses explicitly: `assert(&secret1[0] != &secret2[0]);`.

2. **Sync `scripts/ci/malloc-baseline.txt` with `src/sys/app/app.c`**:
   - Recent changes in `src/sys/app/app.c` shifted lines 3854, 3864, 3986, 3991 down by 3 lines to 3857, 3867, 3989, 3994.
   - Updated the line numbers in `scripts/ci/malloc-baseline.txt` to restore passing status for `scripts/ci/malloc_lint.sh` and the `benchmark report rendering tests` CI workflow.

PR #82, opened 2026-09-13
fix(http2): consume buffered h2c frames before waiting

DPS0340:
## Summary

Related to #25. This PR fixes two h2c stalls. It does not close the HTTP/1 tail-latency investigation.

The HTTP/2 reader can consume plaintext bytes from `conn.read_buf`. Its readiness checks did not check that buffer. After the C1M sniffer consumed a complete request flight, the connection dispatcher waited for new socket input instead of processing those bytes.

The response writer had the same problem with buffered `WINDOW_UPDATE` frames. It could send response headers but wait for flow-control credit that was already in memory.

## Change

- Add one internal readiness predicate that matches `h2_read`.
- For TLS, keep `SSL_pending` as the source of buffered readiness.
- For h2c, check both the replay pointer and its remaining length.
- Use the predicate in the main frame wait and the flow-control receive loop.
- Keep buffer ownership, frame parsing, timeouts and worker defaults unchanged.

## Regression test

`make WERROR=1 test_http2_prebuffer` runs five cases through the real public connection dispatcher and a real nonblocking socketpair:

1. Read both requests from the socket.
2. Split the first HEADERS frame header between replay and socket data.
3. Put both complete requests in the replay buffer.
4. Read a zero-window request and its WINDOW_UPDATE from the socket.
5. Put the same zero-window request and WINDOW_UPDATE in the replay buffer.

The client sends one flight before the server starts. It sends no more bytes until the response verdict is fixed. The test checks response HEADERS, exact DATA bodies and END_STREAM. Cleanup cannot turn a failed verdict into a pass. Setup and checks remain active with NDEBUG.

Observed RED/GREEN sequence on Linux aarch64:

- Original source: socket controls and the split case pass. The fully buffered request case times out after 2.5 seconds with no handler calls.
- Main-wait fix alone: the basic cases pass. The fully buffered flow-control case still times out after response headers, with no DATA.
- Both paths fixed: all five cases pass. The final five-case test also fails against the unchanged original source.

## Verification

- Five normal target groups pass with WERROR=1: new regression, HTTP/2, flow control, HTTPS, and async defer. The async target runs both C1M and classic modes.
- The new regression passes with NDEBUG.
- Five target groups pass with ASan/UBSan and the existing leak-suppression file: new regression, HTTP/2, flow control, HTTPS, and async defer. The new regression also passes with NDEBUG.
- The serial normal suite covers all 82 Makefile targets: 80 pass. Two fail to compile with WERROR=1: `test_middleware_jwt` (array comparison) and `test_webtransport` (unused parameter). Both failures reproduce with unchanged product code in a separate build tree.
- Test wiring passes. The allocator gate reports the same four pre-existing `app.c` line-number mismatches on base and candidate. This change adds no raw allocator calls. PR #81 targets the middleware warning and allocator baseline; this PR does not include those unrelated edits.
- Native GitHub CI, including its conformance jobs, remains pending. This PR stays a draft while those results are unresolved.
- Independent static correctness and security review passed for the exact three-file change.

## Scope

The test models the accept sniffer by reading an exact prefix into the caller-owned replay buffer. It exercises the full connection dispatcher, but not a TCP listener or the complete app/reactor handoff. The fixed-flight HPACK check is not a general HTTP/2 conformance test.

The 2.5-second value is the test deadline, not a measured production latency improvement. This PR makes no HTTP/1 P99.999 claim and does not reduce reactor thread defaults. Long-lived h2c connections still occupy their reactor under the current architecture.


DPS0340:
## CI follow-up at 87fc0f2

All checks have now completed. The HTTP/2 h2spec baseline-diff gate, macOS kqueue job, CodeQL analysis, and benchmark jobs passed. This is not an all-green CI claim:

- ASan/UBSan job [103725193813](https://github.com/c4punks/CWIST/actions/runs/34757834580/job/103725193813) stops at `tests/test_middleware_jwt.c:74:20: error: comparison between two arrays [-Werror=array-compare]`. That compile failure was already reproduced on the unchanged base in the local independent Linux tree; it is not a sanitizer diagnostic from the h2c regression. The targeted local sanitizer tests remain the runtime evidence for that test.
- The report-rendering job fails in its raw-allocator baseline substep, with the same four existing app.c line-number mismatches previously reproduced on the base. Do not describe this as a failing renderer test.

The completed web benchmark artifact (run34757834569, verified head and artifact SHA256) still reports CWIST C1M corrected P99.999 **23.672ms /143,124.63 RPS**, versus Axum **14.368ms /119,980.80 RPS**. Their measured raw logs contain no socket/status error line. This single shared-runner sample is not an A/B of the h2c patch, and the HTTP/1 fixture never exercises the modified h2c readiness paths. It confirms why this PR must not close #25.

A separate manual-only native backend × reactor-thread experiment is being prepared on unchanged runtime source. No runtime default or ownership change is included here.

PR #84, opened 2026-09-13
fix(http): respect header byte length when stringifying responses

DPS0340:
## Summary

Related to #25. This is a small correctness prerequisite found while investigating FIXED-cache population, not a P99.999 performance fix or a cache-enablement change.

`serialize_headers` returns a byte count. Neither its memcpy fast path nor its general path promises a trailing NUL. `cwist_http_stringify_response` discarded that count and passed the stack buffer to the strlen-based string assignment. That can copy unwritten stack bytes or read past the array.

Use the returned length with `cwist_sstring_assign_len`, and return NULL if the string object cannot be created. Body handling and managed-body lifetime remain unchanged. No header-buffer terminator write is needed, including when the returned span fills the buffer.

## Regression

Add `test_http_stringify` to the normal Makefile test list. Six wire cases plus NULL input cover empty/binary bodies, ordinary/managed pointer bodies, keep-alive/close, and custom headers. The test checks exact byte extent, payload bytes, framing and managed cleanup timing without assert-dependent setup. Default headers have an independent expected-byte oracle.

On unchanged dev `7e19b461c5772f4e46e7d58fb96827bbb954cf97`, the normal Linux build runs the regression and fails at `wire->size == header_len + body_len`. The candidate passes the same regression and its NDEBUG build. The original uninitialized-read failure is stack-layout dependent; no claim is made that every original build fails identically.

## Validation

- Linux normal suite: 82 targets reconciled, 80 pass. `test_middleware_jwt` fails on array comparison and `test_webtransport` on an unused parameter; both compiler failures were reproduced on the unchanged base during the preceding investigation.
- New regression passes normally and as an NDEBUG test executable linked to the normal library.
- ASan/UBSan: new regression, HTTP, pipelining, chunked HTTP, HTTPS, BDR, and async defer all pass (existing `tests/lsan.supp` used). Source identities in the isolated build trees match the reviewed candidate.
- Independent exact-three-file SPEC/correctness/security review found no blocker. Test wiring passes.
- The allocator baseline still fails on the same four pre-existing app.c line offsets; no new raw allocation is introduced.
- Native CI has not run for this commit yet.

## Scope limits

This does not restore FIXED caching, change cache eligibility, bypass middleware, or change thread/backend defaults. It does not address the serializer's existing oversized-header behavior, invalid body-mode combinations, or all assignment/append allocation-failure handling. No measured tail-latency improvement is claimed.


DPS0340:
## Exact-head follow-up

Checked updated head `3c78ef6f236c6afa6f7daeb6adfc6f34039b463a` against the previously tested `76f95a39830e65f9188e83339075521f408aad4d`.

The Makefile, HTTP serializer and new stringify test blobs are unchanged. The tree delta consists only of the JWT/WebTransport test fixes from the updated parent. This preserves the scope of the reviewed serializer change; it is not a claim that every new-head check has passed.

New-head ASan/UBSan, h2spec baseline-diff and CodeQL have passed. The report-rendering job actually fails at its raw-allocator baseline substep, not rendering tests: four unchanged app.c allocation/free lines are recorded three lines earlier in the baseline (3854/3864/3986/3991 vs actual 3857/3867/3989/3994). Job: https://github.com/c4punks/CWIST/actions/runs/34761336803/job/103734594906 . Other jobs remain pending at this check; no all-green or merge claim.

The separate safe-cache design is now proposed in #85. It does not activate caching or claim to resolve #25.

DPS0340:
The head changed concurrently to `b34495043e93b30cdf76a827b7abdf212580c524` before my push. Its complete Git tree `7da76b2e20a8f286dbd5b87849dc96251ea7521e` is identical to the reviewed/tested integration candidate, so I did not push a duplicate merge.

- Allocator gate: old baseline RED (four +3 line offsets), current-dev baseline GREEN; wiring PASS.
- Isolated Linux/aarch64 integration: 9/10 normal target groups PASS; 10/10 ASan/UBSan groups PASS, with existing LSan suppressions.
- The remaining normal JWT failure is `-Werror=lto-type-mismatch` for stale metrics stub signatures. Rebuilt pristine dev `65632aed0a30e8aabe1519aa9b706e19685c2b4b` and candidate both reproduce it. This is a separate test-only defect, not a stringify regression. A narrow follow-up is being validated.
- Independent review verified the exact tree preserves both h2 prebuffer and stringify tests and leaves only the original three-file stringify delta against dev.

This is not all-green CI or P99.999 performance evidence. No branch overwrite or PR merge was performed.

PR #86, opened 2026-09-13
fix(test): match JWT metrics stubs to public API

DPS0340:
## Summary

Fix the metrics test doubles in `tests/test_middleware_jwt.c` to match the public API. Include `metrics.h` so later signature drift is diagnosed during compilation, including builds without LTO.

The old stubs omitted the metric ID parameter from `cwist_metric_inc` and used the wrong parameter count/types for `cwist_metric_add`. The registry stub also returned an untyped pointer. The stubs remain no-ops; JWT assertions and runtime code are unchanged. No compiler warning or sanitizer check is disabled.

## Reproduction

On pristine dev `65632aed0a30e8aabe1519aa9b706e19685c2b4b`, Linux/aarch64 GCC with the repository's normal optimization/LTO flags:

```sh
make -j4 WERROR=1 libcwist.a
make WERROR=1 test_middleware_jwt
```

The library builds, but linking the test fails with `-Werror=lto-type-mismatch` for `cwist_metric_inc` and `cwist_metric_add`. The same failure reproduces in the #84 integration tree. Fresh source timestamps forced rebuilds; an earlier stale-build control was discarded.

## Validation

- Linux/aarch64 normal GCC LTO: `test_middleware_jwt`, `test_jwt`, `test_metrics` all pass with `WERROR=1`.
- ASan/UBSan: the same three target groups pass; existing repository LSan suppressions remain enabled.
- Public-header regression guard: correct stubs compile without LTO; a temporary wrong metric-ID type is rejected with a conflicting-types diagnostic. The mutation is not committed.
- Test wiring, allocator baseline gate, and diff whitespace checks pass. The full repository suite was not rerun for this isolated test-only change.

This is a test-only prerequisite found while validating #84 during the #25 investigation, not a P99.999 latency fix. It does not restore FIXED caching or change runtime defaults.


thegoodengineer (commented):
LGTM! Properly aligns the metrics test doubles with the public signatures in metrics.h and resolves -Werror=lto-type-mismatch under LTO builds.

gg582:
LGTM. +11 lines, -3 lines. That's a small patch. IMO we can merge this right now.

PR #91, opened 2026-09-14
feat: prepare v3.5.0 public FIXED cache and reliable benchmarks

### LINKED ISSUE #25: [BUG][Plan/CWIST 3.5] Enhance bad P99.999 latency
## CWIST (3.3) Benchmark Results Summary

| Metric | Concurrency 64 | Concurrency 256 | Concurrency 512 | Notes |
| :--- | :--- | :--- | :--- | :--- |
| **Requests / Second (RPS)** | 96,659 | 93,162 | 88,700 | Steady throughput drop as concurrency rises |
| **P99.999 Latency** | 0.74 ms | 53.77 ms | 645.72 ms | Extreme tail latency spikes under higher load |
| **Average Latency** | 0.14 ms | 3.69 ms | 4.52 ms | Bulk of requests remain within 5 ms |
| **Minimum Latency** | 0.04 ms | 0.04 ms | 0.04 ms | Minimal base overhead regardless of load |

DPS0340:
## Release hold
Do not publish v3.5.0 until the separately developed CVE fixes are included and their associated issue(s) resolved. This condition applies even if this PR and CI pass. The exact dependent issue/PR identifiers still need confirmation; missing identifiers do not waive the hold.

## Summary
- Target v3.5.0, as agreed in https://github.com/c4punks/CWIST/pull/91#issuecomment-5660948894, replacing the proposed v3.4.2 release.
- Implement ADR #85 with explicit PUBLIC_FIXED opt-in, owned representations, conservative admission, process-wide limits, and invalidation. Ineligible requests use ordinary dispatch.
- Integrate #85/#87, including dual-histogram tools and corrected, provenance-bound benchmark reporting. Keep PUBLIC_FIXED as a separate tenth case and preserve existing workload arguments.
- Fix blocking pipeline framing and deferred exchange ownership through middleware unwind, producer completion and inline abort.
- Transfer scheduler and job-queue allocations before publication, including the separately selected kqueue backend. Keep opaque callback arguments caller-owned.
- Finish a coalesced response batch synchronously when shutdown is already observed, rather than parking it on a stopped reactor. This is not a global graceful-shutdown guarantee or an absolute shutdown deadline.
- Package version 3.5.0. Applications must rebuild and relink against matching headers and library.

## Verification
- Baseline `5b128f1178df374ee5bc3db1aba8f921a6af7d68`: Linux ASan/UBSan, source distribution, interop and benchmark jobs passed; macOS C1M TCP failed with a receive timeout.
- Reproduced the macOS failure using the exact CI source archive in an isolated native build. After the shutdown-flush fix, classic/C1M × GC on/off × normal/NDEBUG TCP runs passed (8 runs).
- Common and kqueue job-queue component regressions passed normal/NDEBUG/ASan × GC on/off (12 runs). ASan leak detection was disabled for these native component runs: no leak-proof claim.
- The full native library passed the job handoff target (4 runs) and the TCP target (8 runs) together, with source hashes bound to the integrated candidate.
- Real baseline UAF and scope failures are retained. The portable creator test waits for its callback before stop without relaxing creator/donor exit ordering or GC checks.
- Python CI helper tests: 43 passed. Allocator lint and diff check passed. Actionlint/ShellCheck reports the same pre-existing workflow findings as the baseline, with no new findings.
- Prior source/security reviews and shutdown-flush review passed. Final platform delta is independently reviewed before commit. Exact-head full CI remains required after integration.

## Remaining gates
- Keep draft until exact-head full-library, sanitizer, source archive, native platform, interop and applicable checks pass.
- Keep the separate CVE dependency hold, even after this PR passes.
- Verify the main merge and its checks before any eventual release. Keep existing tags unchanged.
- After release is permitted, verify downloaded assets and remeasure the exact released version with raw artifacts.
- This PR does not close #25 or claim a P99.999 improvement. Failed warmups and older measurements retain their original source/version labels.

Roadmap: `docs/roadmap-v3.5.md`. Related: #25, #85, #87.


gg582:
In X.Y.Z version tag, Z is for urgent patches and critical issue. How about making a roadmap for v3.5 and unify it?

DPS0340:
Agreed, @gg582. Let us target v3.5.0 instead of v3.4.2 and put the public FIXED cache, API changes, and benchmark work into one release roadmap.

The TCP, sanitizer, and CI checks must still pass before release. We will then measure the released version and publish the results. The version change does not mean that #25 is resolved.

DPS0340:
@gg582 unlock draft when you're ready to release :)

PR #102, opened 2026-09-14
fix(grpc): replace atoi with strtol for :status and grpc-status trailer parsing

### LINKED ISSUE #45: [Robustness] Robust Decoding
## Problems

For a good example:

```c
char *DATA= "Josh";
```
Should be refactored like:

```c
char *DATA = "Josh";
char *endptr;

if(endptr != '\0') {
    return NULL;
}
else {
    parse();
}
```

Some behaviors like

```c
char *DATA = "Josh";

void isNum(char *s) {
    // signness.
    int sign = 1;
    if(s[0] == '-') {
        sign  = -1;
    }
    else if (s[0] == '+') {
    }
    else {
            // ...
        }
    }
}
```

Bhumika-1432006:
## What and why

Closes #45

`atoi` offers no way to distinguish a well-formed `"0"` from completely
invalid input — both return `0` with no error signal.  Two places in the
gRPC client header callback were relying on this:

| Field | Old behaviour on bad input | New behaviour |
|---|---|---|
| `:status` | silently becomes `http_status = 0` | becomes `0`, same as before but for a documented reason; values outside 100–599 are explicitly rejected |
| `grpc-status` | silently becomes `CWIST_GRPC_OK` (0), masking the error | becomes `CWIST_GRPC_UNKNOWN` (2) so the caller sees a real failure |

The `grpc-retry-pushback-ms` field right below these two already used
`strtol` correctly; this brings `:status` and `grpc-status` in line with
that established pattern.

## Changes

**`src/net/grpc/grpc_client.c`** — `grpc_client_on_header()`

- `:status`: switched to `strtol` + end-pointer check; accepts `[100, 599]`,
  maps everything else to `0` (existing callers treat `0` as "no status
  received" and surface an error already).
- `grpc-status`: switched to `strtol` + end-pointer check; accepts
  `[0, INT32_MAX]`, maps malformed trailers to `CWIST_GRPC_UNKNOWN` (2)
  instead of `CWIST_GRPC_OK` (0).

No behaviour change for well-formed inputs; only the error path tightens.

## Testing

Existing `test_grpc_channel` exercises the happy path end-to-end and
continues to pass. The two new code paths (bad `:status` and bad
`grpc-status`) are purely defensive — they fire only on malformed server
responses that the existing tests never send, so no test changes are
needed to keep the suite green.


gg582:
Cool. After reading your contributions I could catch that we can specify a scope of anti patterns

PR #103, opened 2026-09-14
fix(app): replace atoi with strtoull in HTTP Range header byte-range parsing

### LINKED ISSUE #45: [Robustness] Robust Decoding

Bhumika-1432006:
## What and why

Closes #45

The static file handler was parsing `Range: bytes=` header values with
`atoi()`, which has two silent failure modes that matter here:

1. **Garbage → 0**: a malformed range like `bytes=abc-def` becomes
   `range_start = 0`, silently serving from the beginning of the file
   rather than returning `416 Range Not Satisfiable`.
2. **Overflow on large files**: byte positions above `INT_MAX` (~2 GB)
   wrap silently.  Range headers on large static files are a normal use
   case.

Both are exactly the kind of robust-decoding failure described in #45.

## Changes

**`src/sys/app/app.c`** — `cwist_static_handler()` Range parsing block

| Case | Old behaviour | New behaviour |
|---|---|---|
| Suffix range `bytes=-abc` | `suffix = 0`, serves last 0 bytes | `range_valid = false` → 416 |
| Start position `bytes=abc-` | `range_start = 0`, serves from byte 0 | `range_valid = false` → 416 |
| Value > INT_MAX | silent truncation | handled correctly by `unsigned long long` |
| Well-formed range | unchanged | unchanged |

The `range_valid` flag was already declared in the original code but was
only used as `(void)range_valid;` (suppressing the unused-variable
warning without actually checking it).  It is now wired into the
`416` guard alongside the existing out-of-bounds check.

## Testing

The existing `test_static_and_range.c` exercises valid range requests
and continues to pass.  The new error paths (non-numeric range tokens)
were not tested before this change; they now return 416 consistently
with RFC 9110 §14.1.2.


gg582:
Cool. That's basically about HTTP/416 not happening when range is invalid

PR #104, opened 2026-09-14
fix(http2): replace atoi/atol with strtol in HTTP/2 env-config reader functions

### LINKED ISSUE #45: [Robustness] Robust Decoding

Bhumika-1432006:
## What and why

Closes #45

Five env-var reader functions in `src/net/http/http2.c` were using
`atoi` / `atol` to parse configuration values, each with a
double-call anti-pattern:

```c
timeout_ms = (env && atoi(env) > 0) ? atoi(env) : DEFAULT;
```

Problems:
- **Calls the parser twice** on the same string — unnecessary work and
  technically a TOCTOU if the env var could change (unlikely here, but
  worth fixing on principle).
- **Silent failure on garbage input**: `atoi("200abc")` returns `200`
  with no indication that the trailing characters were ignored.
  A misconfigured env var like `CWIST_HTTP2_IDLE_TIMEOUT_MS=30s` would
  silently apply `30` instead of the intended `30000`, and the operator
  would have no idea.

## Changes

**`src/net/http/http2.c`**

| Function | Env var |
|---|---|
| `h2_idle_timeout_ms` | `CWIST_HTTP2_IDLE_TIMEOUT_MS` |
| `h2_goaway_grace_ms` | `CWIST_HTTP2_GOAWAY_GRACE_MS` |
| `h2_max_rst_burst` | `CWIST_HTTP2_MAX_RST_BURST` |
| `h2_max_rst_rate` | `CWIST_HTTP2_MAX_RST_RATE` |
| `h2_batch_threshold` | `CWIST_H2_BATCH_BYTES` |

All five now:
1. Call `strtol` once and capture the value in a local variable.
2. Check `end != env && *end == '\0'` to reject any trailing garbage.
3. Fall back to the compile-time default on any parse failure, same as
   an unset variable.

No behaviour change for well-formed numeric strings.

## Testing

All HTTP/2 tests (`test_cwist.c`, `test_compress.c`, etc.) continue to
pass.  The new rejection paths only fire on malformed env var values
that the test suite never sets.


gg582:
Nice catch. Thanks for your contribution!

Issue #65, opened 2026-09-13
Full-GC mode (cwist_full_gc(true)) has unprofiled ~86% per-call malloc/free overhead


gg582:
## Summary

`cwist_full_gc(true)` mode has real, measured overhead on every
`malloc`/`calloc`/`realloc`/`free` call — both through `cwist_alloc()` and
through the newly-shipped `CWIST_INTERCEPT_MALLOC` shim (#61) — and that
overhead has not been profiled or reduced. This tracks improving it.

## What's measured today

`tests/bench_malloc_intercept.c` (single-threaded, mixed 16–512 byte
allocations, occasional `realloc`), one machine:

| configuration | ns/op | vs. baseline |
|---|---:|---:|
| baseline (plain `malloc`/`free`, no shim) | ~11.8 | — |
| shim active, full-GC off (default) | ~12.2 | +3% |
| shim active, full-GC **on** | ~21.9 | **+86%** |

The "off" number is fine and expected (one relaxed atomic load plus one
extra call). The "on" number — roughly **2x per-call cost** — is the open
question. It comes from `cwist_gc_scope_track()`/`cwist_gc_scope_untrack()`
maintaining a per-thread pending-sweep list on every allocation and
release (`src/core/mem/gc.c`).

## What is not known yet

- Where the ~10ns of extra per-call cost actually goes: list insertion
  itself, an atomic/memory-ordering cost in the tracking structure,
  allocation of the tracking node, or something else. No profiling
  (`perf`, or even manual instrumentation) has been done — the number
  above is only an aggregate measurement, not a breakdown.
- Whether the cost scales with the number of objects currently pending
  sweep (i.e. is the pending-sweep list a data structure with O(1)
  insert/remove, or does cost grow as more objects accumulate between
  flushes?).
- Concurrent-load behavior: the benchmark above is single-threaded by
  design (isolates per-call overhead, not lock contention) — there is no
  measurement yet of the cost under many threads allocating/freeing
  concurrently, which is the realistic C1M/classic-pool deployment shape.
- Whether the same ~86% figure holds for `cwist_alloc()`/`cwist_free()`
  directly (it should, since the shim reuses the exact same tracking
  calls) — this hasn't been separately re-measured after #61, only
  inferred.

## Suggested next steps

1. Profile a tight alloc/free loop with full-GC on (`perf record`/`perf
   stat`, or manual `rdtsc`-style instrumentation around
[...]

PR #105, opened 2026-09-14
perf(gc): add bench_malloc_intercept.c — concurrent full-GC tracking overhead harness (issue #65)

Bhumika-1432006:
## What and why

Refs #65

Issue #65 measured ~86% per-call overhead when `cwist_full_gc(true)` is
on and identified three suggested next steps:

1. Profile a tight alloc/free loop with full-GC on to find where the ~10 ns goes.
2. **Re-run under concurrent multi-threaded load** to check whether the
   pending-sweep list is a contention bottleneck — not just a per-call cost.
3. Evaluate mitigations after the above.

This PR addresses step 2 by adding the missing benchmark harness that was
referenced in the issue but never committed to the tree.

## The file: `tests/bench_malloc_intercept.c`

Measures three configurations across two concurrency shapes:

| Config | Description |
|---|---|
| `baseline` | Plain `malloc`/`free`, no CWIST shim |
| `shim-off` | `cwist_alloc`/`cwist_free`, `full_gc` disabled |
| `shim-on` | `cwist_alloc`/`cwist_free`, `full_gc` enabled |

**Single-threaded** — reproduces the original measurements from the issue
to confirm the numbers on the target machine.

**Multi-threaded** (8 threads, each doing 1 M iterations) — reveals whether
the per-thread pending-sweep list overhead is truly per-thread (constant
regardless of thread count) or whether the threads share state and contend
(overhead grows with thread count). That distinction matters for choosing the
right mitigation in step 3.

The allocation mix is intentionally representative of a real handler:
small metadata structs (16–64 bytes), mid-sized buffers (128–512 bytes),
and a `realloc` every 20 iterations.

## How to run

```sh
cc -O2 -std=c17 -pthread \
    -I include -I lib/libttak/include \
    tests/bench_malloc_intercept.c \
    src/core/mem/alloc.c src/core/mem/gc.c src/core/mem/arena.c \
    -L lib/libttak -lttak -o bench_malloc_intercept
./bench_malloc_intercept
```

Results should be shared in issue #65 for the next step (profiling /
mitigation design).


gg582:
Hi, we have our official discord.

Please join here via https://discord.gg/6F8HDmNAPg

PR #106, opened 2026-09-14
docs(gc): document full_gc per-call overhead caveat and point to issue #65

Bhumika-1432006:
## What and why

Refs #65

`docs/GC.md` is the design document for CWIST's full-GC mode.  It explains
what the mode does, how to enable it, and what the semantics are — but it
had no mention of the known ~86% per-call overhead that `cwist_full_gc(true)`
currently introduces on every `cwist_alloc` / `cwist_free` pair.

A developer who reads the docs, enables full-GC mode on a high-throughput
service, and then wonders why their allocation-heavy handlers got slower has
no pointer to the existing investigation or benchmark harness.

## Changes

**`docs/GC.md`** — new "Known performance caveat" section at the bottom:

- States the measured overhead (~10 ns/op, ~86% over baseline on one machine).
- Links to issue #65 where the root cause and concurrent-load behaviour are
  being investigated.
- Points to `tests/bench_malloc_intercept.c` (added in the companion PR #105)
  for anyone who wants to reproduce or refine the measurement.
- Gives practical guidance: profile the hot path before opting into full-GC
  mode on a high-throughput service.

Documentation only; no behaviour change.


gg582:
@Bhumika-1432006 Hi, I just noticed that we've merged pull requests to main; next time, please merge this into `dev` branch. Thanks!

PR #109, opened 2026-09-14
fix(mem): cwist_realloc() didn't update full-GC pending-sweep tracking

gg582:
Found while verifying #105's newly-merged `bench_malloc_intercept.c`: running it crashes with `free(): double free detected in tcache 2`.

## Root cause

`cwist_alloc()` tracks its result on the calling thread's full-GC pending-sweep list, but `cwist_realloc()` called libc `realloc()` directly and never updated that list. `realloc()` is free to move the block to a new address; when it does, the pending list keeps pointing at the stale original address — memory that's already back in general circulation. At thread exit, the sweep dutifully `free()`s that stale address, corrupting whatever tcache slot it landed in by then.

Backtrace (gdb, `-O0` build):
```
#5  malloc_printerr (..."free(): double free detected in tcache 2")
#10 cwist_free (ptr=...) at src/core/mem/alloc.c:400
#11 cwist_ebr_free_cb (...) at src/core/mem/gc.c:111
#12 ttak_epoch_reclaim () at .../libttak/src/mem/epoch.c:572
#14 cwist_gc_pipeline_tick () at src/core/mem/gc.c:128
#15 cwist_gc_pending_destroy (...) at src/core/mem/gc.c:299   <- thread-exit sweep
#16 __nptl_deallocate_tsd ()
```

## Fix

`cwist_realloc_shim()` (the `CWIST_INTERCEPT_MALLOC` header shim's `realloc()` replacement, `dev`-only) already implements the correct untrack-old/track-new rule. This brings plain `cwist_realloc()` in line: untrack the incoming pointer (if tracked) before calling `realloc()`/the owner-guarded equivalent, track whichever address it returns, and restore the original tracking if the (re)allocation fails (`realloc()` leaves the original block valid on failure, so untracking it unconditionally would itself leak the safety net).

## Verification

- Reproduced the crash on the pre-fix build via `bench_malloc_intercept`, confirmed under both gdb and valgrind; both clean after the fix.
- Full `make test` passes with this fix (running now as final confirmation before merge — will update this PR).
- Real production call site affected: `src/net/grpc/grpc_client.c`'s message-buffer growth path (`cwist_realloc(block, len + cont.len)`) — would have silently corrupted memory under full-GC mode + a growing gRPC message.

Refs #65. Same fix already landed on `dev` (independently, since `dev` hit this by way of a cherry-pick from #105/#106).


gg582:
Confirmed: full `make test` on `main` with this fix — exit code 0, all 31 test binaries pass, no double-free/corruption/assertion failures anywhere in the suite.

Issue #97, opened 2026-09-14
[CVE/http2] Made You Reset: unthrottled server-initiated RST_STREAM


gg582:
## Summary

CWIST's HTTP/2 layer already has explicit, well-targeted mitigations for **Rapid Reset (CVE-2023-44487)** and the **CONTINUATION Flood** class (CVE-2024-27983 and siblings) — both cited by name in `src/net/http2.c` comments, both token-bucket rate-limited. What's missing is a rate limit on **server-initiated** `RST_STREAM`, which is exactly the gap the newer **"Made You Reset"** attack class exploits.

## The gap, with code

`h2_send_rst_stream()` (`src/net/http/http2.c:964-971`) has no budget check — it's a pure frame-writer, called unconditionally by every caller:

```c
static int h2_send_rst_stream(h2_conn *hc, uint32_t stream_id, uint32_t error_code) {
    /* ...builds the RST_STREAM payload, calls h2_write_frame()... */
}
```

Call sites: `2887`, `2914`, `2945`, `2964`, `3051`. The concrete attack path is `2884-2890` (`h2_request_stream_create()`):

```c
if (hc->active_streams >= CWIST_HTTP2_MAX_CONCURRENT_STREAMS) {   // 2884, cap = 100
    h2_send_rst_stream(hc, stream_id, H2_ERR_REFUSED_STREAM);      // 2887, unthrottled
    *refused = true;
    return NULL;
}
```

Once a connection has 100 streams open (trivial: open 100 real streams and simply never close them — nothing here requires the client to send its own `RST_STREAM`, so the existing client-RST token bucket at `499-511` never engages), **every subsequent `HEADERS` frame for a new stream ID costs the server one full frame-parse + HPACK-decode + `RST_STREAM` write**, at whatever rate the client can push minimal `HEADERS` frames — which, with HPACK static/dynamic table reuse, is cheap for the attacker and asymmetric against the server. This is the same cost-asymmetry shape as Rapid Reset, just triggered by exceeding the concurrency cap instead of by client-sent resets, which is exactly what the "Made You Reset" research describes as sidestepping a client-RST-only limiter.

## Suggested fix

[...]

PR #110, opened 2026-09-14
fix(http2): gate outbound RST_STREAM with token bucket (Made You Reset / #97)

Bhumika-1432006:
## Problem

Issue #97 ("Made You Reset"): a client that fills the 100-stream concurrency cap and floods HEADERS frames forces the server to emit an unlimited stream of RST_STREAM REFUSED_STREAM frames — one per incoming HEADERS — with zero cost to the attacker and unbounded write amplification on the server side.

The existing inbound token bucket (added for CVE-2023-44487 / Rapid Reset) only limits *client*-sent RST_STREAM frames; there was no corresponding guard on the server's own RST_STREAM output.

## Fix

Add a token bucket for **outbound** server-initiated RST_STREAM that mirrors the existing inbound budget.

### New state in `h2_conn`

```c
uint32_t out_rst_budget;        /* outbound RST_STREAM credits remaining */
uint64_t out_rst_last_refill_ms; /* last refill timestamp */
```

Both are initialized to `h2_max_rst_burst()` in `h2_conn_init()`, the same starting value as the inbound bucket.

### New helper: `h2_conn_consume_out_rst_budget()`

Same token-bucket refill logic as the inbound `h2_conn_consume_rst_budget()`:

- Full refill to `h2_max_rst_burst()` every ≥1 s
- Proportional top-up via `h2_max_rst_rate()` for sub-second intervals
- Returns `false` when depleted

### Updated `h2_send_rst_stream()`

Before writing the RST_STREAM frame the function now calls `h2_conn_consume_out_rst_budget()`. When the budget is exhausted it logs a warning, emits `GOAWAY ENHANCE_YOUR_CALM`, and returns `-2` (connection-level error). All callers propagate this as a connection teardown so the attack vector is shut immediately instead of the connection remaining open as an RST factory.

### Callers updated

| Site | Change |
|------|--------|
| `h2_request_stream_create()` | Added `bool *goaway_sent` out-param so callers distinguish budget-exhaustion from a normal stream refusal |
| `h2_decode_stream_headers()` | Propagates `-2` from RST as `-1` (connection error) |
| `h2_begin_headers()` (two sites) | Propagates `-2` directly |
| Continuation handler | Uses the new `goaway_sent` flag |

### No new configuration surface

The budget reuses the existing `CWIST_HTTP2_MAX_RST_BURST` / `CWIST_HTTP2_MAX_RST_RATE` environment variables, so operators who already tune the inbound limits get symmetric outbound protection automatically.

## Testing

The attack scenario is: open 100 streams (saturating `CWIST_HTTP2_MAX_CONCURRENT_STREAMS`), then flood HEADERS frames. Without this patch the server writes one RST_STREAM per frame with no upper bound. With this patch the server writes at most `MAX_RST_BURST` RST frames in the first window, then issues GOAWAY and drops the connection.


gg582:

Closing this in favor of https://github.com/c4punks/CWIST/pull/121.

Issue #98, opened 2026-09-14
[CVE/tls] Process-lifetime TLS session-ticket key never rotates


gg582:
## Summary

CWIST shares one TLS session-ticket encryption key (STEK) across every worker process for the life of the process, so a resumed session/0-RTT-eligible ticket works no matter which pre-forked worker accepts the reconnection — a real, deliberate, already-documented design choice (`src/net/http/https.c:757-758`):

> "the key lives for the process lifetime (no rotation), which is the standard pre-fork server model (same as a fixed nginx ticket key)."

This issue isn't disputing that trade-off — it's tracking the concrete downside that comes with it, since it lines up with a known resumption-replay risk class: the longer a STEK lives, the longer the window an intercepted/leaked ticket (or a replayed one, for implementations that support 0-RTT — CWIST doesn't, see below) stays valid across restarts-free uptime.

## Where it lives

`src/net/http/https.c`:
- `760-764`: `cwist_tls_ticket_key` (name/AES-256-CBC key/HMAC-SHA-256 key).
- `803-817` (`cwist_tls_setup_shared_ticket_key`): generated once via `RAND_bytes()` at context-init time, installed via `SSL_CTX_set_tlsext_ticket_key_cb()`, never regenerated afterward.
- `773-796` (`cwist_tls_ticket_key_cb`): the encrypt/decrypt callback BoringSSL invokes per handshake — always reads the same process-lifetime key.

Worth noting explicitly since it narrows the actual exposure here: **0-RTT/early data is not implemented anywhere in this codebase** (no `SSL_CTX_set_max_early_data` or equivalent) — so the specific 0-RTT replay attack (an idempotent-looking request replayed within the ticket's early-data window) doesn't apply. What remains is the more general point that a long-lived, process-wide STEK is a bigger blast radius if it ever leaks (memory disclosure, core dump, etc.) than a rotated one would be — every ticket ever issued by that process stays decryptable for as long as the key does.

## Suggested direction (not a required fix — a tracking issue for the trade-off)

- Periodic in-process key rotation (e.g. generate a new key on a timer, keep the previous one around just long enough to decrypt tickets issued under it, same overlap window nginx's own `ssl_session_ticket_key` rotation guidance uses).
[...]

PR #111, opened 2026-09-14
docs(https): document STEK process-lifetime trade-off and rotation path (closes #98)

Bhumika-1432006:
## Summary

Issue #98 tracks the fact that CWIST's shared session-ticket encryption key (STEK) lives for the entire process lifetime with no rotation. The issue explicitly accepts a documentation-only resolution: *"downgrading this to a documented decision is a legitimate resolution too."*

This PR takes that path. The existing two-line note at `https.c:757-758` was accurate but minimal. This expands it to a complete decision record.

## What the new comment covers

**Why rotation is non-trivial here**
In a pre-fork model the key must be visible across worker processes before the fork. Rotation after the fork requires either a shared `mmap` region with a lock or a control socket from the parent — non-trivial complexity the issue acknowledges.

**0-RTT does not apply**
CWIST never calls `SSL_CTX_set_max_early_data()` or equivalent, so early data / 0-RTT is disabled at the library level. The 0-RTT replay attack vector does not exist in this codebase.

**What risk does remain**
If the STEK leaks (memory disclosure, core dump, cold-boot) an attacker can decrypt any session ticket issued by this process for its remaining uptime. Rotation would cap that window.

**Three mitigations short of full rotation**
1. ECDHE cipher-suite preference (already the default via BoringSSL's ordering) — PFS means bulk traffic keys are independent of the STEK even if it leaks.
2. Short restart cadence (e.g. daily) — a new process generates a fresh STEK.
3. Disable ticket resumption entirely with `SSL_OP_NO_TICKET` to force full handshakes.

**How to implement rotation if it is ever added**
Step-by-step sketch: generate new key on timer, keep one previous key for the overlap window, atomic pointer swap or RW-lock in `cwist_tls_ticket_key_cb()`.

## Files changed

- `src/net/http/https.c`: block comment above `cwist_tls_ticket_key` expanded (no code changes).


gg582:
Hi, I've mentioned already but you'd better make a pull request to `dev`

gg582:

Closing this in favor of https://github.com/c4punks/CWIST/pull/121.

PR #114, opened 2026-09-14
fix(http): replace atoi/atol with strtol in env-var readers (issue #45)

Bhumika-1432006:
## What

Seven call sites in `src/net/http/http.c` use `atoi()`/`atol()` to parse process-configuration env vars. These functions silently accept leading/trailing junk and overflow, so a value like `CWIST_WORKER_THREADS=8x` would silently parse as `8` and `CWIST_POOL_IDLE_TIMEOUT_MS=2000ms` as `2000` — no error, no log line, just a silently misread config.

## Fix

Replace every site with `strtol()` and end-pointer validation (`end != env && *end == '\0'`). An invalid string falls back to the same default as before — no behaviour change for well-formed values, and incorrect values now reliably use the safe default rather than a partial parse.

| Env var | Function |
|---------|---------|
| `CWIST_WORKER_THREADS` | `get_optimal_thread_count()` |
| `CWIST_WORKERS` | `get_optimal_thread_count()` |
| `CWIST_POOL_IDLE_TIMEOUT_MS` | `http_dynamic_worker_thread()` |
| `CWIST_POOL_PREWARM` | `http_pool_init()` |
| `CWIST_HTTP_KEEP_ALIVE_TIMEOUT` | `cwist_http_keep_alive_timeout_sec()` |
| `CWIST_USE_TCP_CORK` | `cwist_tcp_cork_enabled()` |
| `CWIST_TCP_CORK_BURST` | `cwist_tcp_cork_burst()` |

Mirrors the same pattern already applied to `http2.c` env readers in #104 and to the HTTP Range header in #103.


gg582:

PR #115, opened 2026-09-14
fix(redis): replace atoll with strtoll + add bulk-string size cap (issue #45)

Bhumika-1432006:
## What

`src/net/redis/cwist_redis.c`'s RESP2 parser calls `atoll()` to read the bulk-string byte count and array element count from lines sent by the Redis server. `atoll()` accepts trailing garbage and returns 0 for non-numeric input with no way to distinguish success from failure, which matters when the data source is the network.

## Why it matters

A rogue or compromised Redis server can send:

- `$2147483647\r\n` — `atoll` parses as 2 GiB, `cwist_alloc(2GiB + 1)` is called before reading a single byte → OOM / process kill.
- `$abc\r\n` — `atoll` returns 0, caller treats it as an empty string → silent data corruption.

The array-count path has the same non-numeric issue; a huge positive count drives an unbounded read loop.

## Fix

Three call sites replaced:

| Site | Old | New |
|------|-----|-----|
| `'$'` bulk-string length in `read_reply()` | `atoll(line+1)` | `strtoll` + end-pointer check; error on junk |
| `'*'` array count in `read_reply()` | `atoll(line+1)` | `strtoll` + reject non-numeric or negative |
| `'*'` array count in subscribe loop | `atoll(line+1)` | `strtoll` + skip malformed line |

Also adds `CWIST_REDIS_MAX_BULK_BYTES` (512 MiB — Redis's own protocol cap) checked **before** `cwist_alloc()` to bound allocations from a misbehaving server.

No behaviour change for well-formed RESP2 responses.


gg582:
Hi, I mentioned earlier but you'd better make a pull request to `dev` branch.

gg582:

PR #116, opened 2026-09-14
fix(https,app): replace atol/atoi with strtol in env-var readers (issue #45)

Bhumika-1432006:
## What

Two remaining `atoi`/`atol` call sites that weren't covered by the earlier #45 PRs.

### `src/net/http/https.c` — `CWIST_HTTPS_HS_SHARDS`

`https_hs_shepherd_start()` reads the handshake-shard count with `atol(env)`.  A value like `"4x"` silently parses as `4`; an empty string returns `0`, which falls outside the `[1, MAX_SHARDS]` guard (keeps the bug benign but the parsing is still wrong).

Fixed with `strtol()` + end-pointer check; only a clean decimal integer within the valid range is accepted.

### `src/sys/app/app.c` — `CWIST_WORKERS`

`cwist_app_listen()` calls `atoi(workers_env)` for the non-`"auto"` path.  A value like `"8x"` silently becomes `8`.  The downstream `if (workers < 1) workers = 1` guard catches a `0` return from fully-junk input, but it doesn't catch a partial parse like `"4x"`.

Fixed with `strtol()` + full validation (`end != env && *end == '\0' && v >= 1 && v <= INT_MAX`); a junk value now falls back to `1` worker instead of silently using the partial parse.


gg582:

PR #117, opened 2026-09-14
fix(websocket): add per-frame payload size cap before cwist_alloc

### LINKED ISSUE #119: [bug/websocket] No per-frame payload size limit — OOM DoS via oversized frame
## Summary

`cwist_websocket_receive()` in `src/net/websocket/websocket.c` reads the frame payload length from the two-byte header (RFC 6455 allows up to 2^63-1 bytes in the 64-bit extended length field) and calls `cwist_alloc(payload_len + 1)` with no upper-bound check.

## Attack scenario

Any connected WebSocket client — including an unauthenticated one if the upgrade endpoint has no auth guard — can send a single frame header advertising a 4 GiB (or larger) payload:

```
0x82 0xff [4-byte mask] [8-byte length = 0x0000000100000000]
```

The server calls `cwist_alloc(4 GiB + 1)` before reading a single byte of payload. On most deployments this:
- triggers the OOM-killer, taking out the worker process, OR
- causes `calloc` to return NULL and the connection to drop

Either way the attacker expends ~60 bytes and the server expends a syscall and a potential process kill.

## Location

`src/net/websocket/websocket.c` — the `cwist_alloc(payload_len + 1)` call in `cwist_websocket_receive()`, after the masking-key read and before `read_exact(ws->fd, payload, payload_len)`.

## Suggested fix

Check `payload_len` against a configurable maximum (e.g. 16 MiB default) immediately after resolving the extended length:

```c
#define CWIST_WS_MAX_PAYLOAD_BYTES ((uint64_t)(16u * 1024u * 1024u))

if (payload_len > CWIST_WS_MAX_PAYLOAD_BYTES) return NULL;
```

Applications that legitimately need larger frames should use WebSocket message fragmentation (RFC 6455 §5.4).

Bhumika-1432006:
## Problem

`cwist_websocket_receive()` reads the payload length from the WebSocket frame header (a `uint64_t` that can represent up to ~9 exabytes per RFC 6455) and calls `cwist_alloc(payload_len + 1)` with no upper-bound check. Any connected WebSocket client can send:

```
0x82 0xff [4-byte mask] [8-byte length = 0x0000000100000000]
```

The server calls `cwist_alloc(4 GiB + 1)` before reading a single byte of payload. On most deployments this triggers the OOM-killer or causes `calloc` to return NULL, disrupting the server at the cost of a single TCP connection.

Closes #119.

## Fix

Add `CWIST_WS_MAX_PAYLOAD_BYTES` (16 MiB) checked immediately after reading the masking key:

```c
if (payload_len > CWIST_WS_MAX_PAYLOAD_BYTES) return NULL;
```

Frames that exceed the cap are rejected the same way a read error or an unmasked frame is — the function returns `NULL` and the caller is expected to close the connection.

**16 MiB** is a sensible default for a framework targeting API and real-time messaging use cases. Applications transferring larger payloads should use WebSocket message fragmentation (RFC 6455 §5.4): multiple `FIN=0` continuation frames followed by a `FIN=1` final frame, each within the cap.

Frames up to 16 MiB are unaffected. The WebSocket echo example and the test suite operate with payloads well below this limit.


gg582:
Hi, I mentioned earlier but you'd better make a pull request to dev branch.

gg582:

Closing this in favor of https://github.com/c4punks/CWIST/pull/127.

PR #128, opened 2026-09-14
fix(websocket): echo CLOSE frame back to peer on receive (RFC 6455 §5.5.1)

### LINKED ISSUE #120: [bug/websocket] CLOSE frame not echoed back — RFC 6455 §5.5.1 violation
## Summary

When `cwist_websocket_receive()` receives a WebSocket CLOSE frame (`opcode == CWIST_WS_FRAME_CLOSE`), it sets `ws->is_closed = true` and returns the frame to the caller but **does not send a CLOSE frame back to the client**. This violates RFC 6455 §5.5.1.

## RFC requirement

> If an endpoint receives a Close frame and did not previously send a Close frame, the endpoint MUST send a Close frame in response.

The reciprocal CLOSE is the orderly shutdown signal. Without it, the peer's WebSocket implementation (browser or library) cannot distinguish a clean close from a dropped connection, and many implementations will emit an error event or retry rather than treating the session as cleanly terminated.

## Current behaviour

`src/net/websocket/websocket.c` around line 187:
```c
if (opcode == CWIST_WS_FRAME_CLOSE) {
    ws->is_closed = true;
}
```

No CLOSE response is sent. The socket is left open until the caller destroys the `cwist_websocket *`.

## Suggested fix

Before setting `is_closed`, send a CLOSE frame back using `cwist_websocket_send()` with the CLOSE opcode and, optionally, mirroring the status code from the received frame's payload:

```c
if (opcode == CWIST_WS_FRAME_CLOSE) {
    cwist_websocket_send(ws, CWIST_WS_FRAME_CLOSE, payload, payload_len < 2 ? 0 : 2);
    ws->is_closed = true;
}
```

A guard against double-send (if the application also called `cwist_websocket_close()`) may be needed.

Bhumika-1432006:
## Problem

RFC 6455 §5.5.1:

> If an endpoint receives a Close frame and did not previously send a Close frame, the endpoint MUST send a Close frame in response.

`cwist_websocket_receive()` was silently setting `ws->is_closed = true` when it received a CLOSE frame but never sent the reciprocal CLOSE frame. Browser WebSocket implementations and well-conforming libraries treat the absence of a CLOSE echo as an abnormal closure and fire an `onerror` / error event rather than a clean `onclose`.

Closes #120.

## Fix

One added call to `cwist_websocket_send()` before setting `is_closed`:

```c
cwist_websocket_send(ws, CWIST_WS_FRAME_CLOSE,
                     payload, (payload_len >= 2) ? 2 : 0);
ws->is_closed = true;
```

The first two bytes of the received CLOSE payload are the status code (RFC 6455 §5.5.1 says to echo it); if the peer sent an empty CLOSE body an empty CLOSE is sent back instead.

## No API change

Callers that were checking `frame->opcode == CWIST_WS_FRAME_CLOSE` already have the right pattern; the CLOSE frame is still returned to them so they can see the peer's status code. The only change is that the socket now gets a proper CLOSE before `is_closed` flips.


gg582:
Landed via #133 (bundled with the fragmentation reassembly from #131, cherry-picked with authorship preserved, merged). Thanks!

Bhumika-1432006:

PR #131, opened 2026-09-14
fix(websocket): reassemble fragmented messages per RFC 6455 §5.4

### LINKED ISSUE #122: [bug/websocket] Fragmented messages not reassembled — fin flag ignored by cwist_websocket_receive()
## Summary

RFC 6455 §5.4 allows a WebSocket message to be split across multiple frames:
- the first frame has `FIN=0` and a non-CONTINUATION opcode (text or binary)
- intermediate frames have `FIN=0` and `opcode=0` (CONTINUATION)
- the final frame has `FIN=1` and `opcode=0`

`cwist_websocket_receive()` reads and returns each frame individually. The `fin` field is set on the returned `cwist_ws_frame` struct but the function never buffers and reassembles a multi-frame message. This means:

1. An application calling `cwist_websocket_receive()` in a loop will receive individual fragments rather than complete messages.
2. A CONTINUATION frame with `opcode=0` will be returned as-is; the caller has no way to identify which in-progress message it belongs to.
3. A peer that chooses to fragment — either for flow control, or because it is a well-conforming library — will have its messages silently split.

## Location

`src/net/websocket/websocket.c` — `cwist_websocket_receive()`. The `fin` variable is read from the frame header but is only stored in `frame->fin`; no reassembly state is kept on `cwist_websocket`.

## Impact

Most browser WebSocket implementations do not fragment small messages in practice, which is why this bug has likely gone unnoticed. However:
- Libraries like `websocket-client` (Python) do fragment large messages.
- The spec does not forbid fragmentation even for small frames, so any spec-compliant peer could trigger this.
- Binary protocols that stream large blobs will arrive fragmented and be presented as partial payloads to the application.

## Suggested fix

Add a reassembly buffer to `cwist_websocket`:
- On a frame with `FIN=0`: accumulate the payload into a per-connection buffer.
- On the final frame (`FIN=1`, `opcode=CONTINUATION`): append its payload, set the combined opcode on the returned frame, and reset the buffer.
- While accumulating, enforce a maximum total assembled-message size (tied to or independent of `CWIST_WS_MAX_PAYLOAD_BYTES`).

Bhumika-1432006:
## Problem

`cwist_websocket_receive()` returned each WebSocket frame individually, leaving the `fin` flag as raw metadata for the caller. RFC 6455 §5.4 allows a logical message to be split across any number of frames:

```
first frame:   FIN=0, opcode=TEXT or BINARY
middle frames: FIN=0, opcode=CONTINUATION (0x0)
last frame:    FIN=1, opcode=CONTINUATION
```

A conforming peer that chose to fragment a large message would have each fragment delivered as a separate incomplete frame, breaking every application that reads full messages from `cwist_websocket_receive()`.

Closes #122.

## Fix

**New state on `cwist_websocket`:**
```c
uint8_t          *frag_buf;
size_t            frag_len;
size_t            frag_cap;
cwist_ws_opcode_t frag_opcode;
```

**Inside `cwist_websocket_receive()`:**
- Frames with `FIN=0` are appended to `frag_buf` and the function loops to read the next frame (no return to caller yet).
- The final `FIN=1 CONTINUATION` frame flushes the buffer; the assembled payload is returned as a single frame with the original opcode (`TEXT` or `BINARY`).
- Single-frame messages (`FIN=1` with a non-CONTINUATION opcode) take the fast path through the existing logic — no overhead for the common case.
- Control frames (`CLOSE`, `PING`, `PONG`) are always `FIN=1` per spec and are delivered immediately even while a data fragmentation is in progress.

**`cwist_websocket_destroy()`** now frees `frag_buf` to prevent a leak when a session is torn down mid-reassembly.

## No API change

`cwist_websocket_receive()` signature is unchanged. Callers that worked with single-frame messages see identical behaviour; multi-fragment messages now arrive as a single assembled frame instead of broken parts.


gg582:
Hi, I've already mentioned that you should not make a pull requests to `main`.

gg582:
Landed via #133 (bundled with the CLOSE-echo fix from #128, cherry-picked with authorship preserved, merged). Thanks!

Bhumika-1432006:

PR #132, opened 2026-09-14
fix(http): move HSTS out of generic security-headers helper

### LINKED ISSUE #123: [bug/http] HSTS header emitted on plain-HTTP responses — RFC 6797 §7.2 violation
## Summary

`cwist_http_response_add_security_headers()` in `src/net/http/http.c` unconditionally appends `Strict-Transport-Security: max-age=31536000; includeSubDomains` to every HTTP response, regardless of whether the connection is over TLS.

RFC 6797 §7.2 states:

> An HSTS Host MUST NOT include the STS header field in HTTP responses conveyed over non-secure transport.

Sending HSTS over plain HTTP is a no-op in compliant browsers (they ignore it) — but it is still a spec violation that can cause issues with:
- Strict HSTS validators and security scanners that flag the header as a misconfiguration.
- Reverse proxies or test harnesses that inspect headers and interpret an HSTS header on HTTP as a configuration error.
- Spec-enforcing HTTP clients that may reject responses with HSTS on non-TLS connections.

## Location

`src/net/http/http.c` — inside `cwist_http_response_add_security_headers()`:

```c
if (!cwist_http_header_get(res->headers, "Strict-Transport-Security")) {
    cwist_http_header_add_static(&res->headers, arena,
        "Strict-Transport-Security", "max-age=31536000; includeSubDomains");
}
```

The function has no visibility into whether the response is being sent over TLS.

## Suggested fix

Two options:

1. **Pass an `is_tls` flag** to `cwist_http_response_add_security_headers()` and only emit HSTS when `is_tls` is true.
2. **Remove HSTS from `cwist_http_response_add_security_headers()`** and add it only in the HTTPS-specific response path (e.g. in the `cwist_https_*` handler layer where TLS is guaranteed).

Option 2 is cleaner because it keeps the security-header function transport-agnostic and avoids a signature change.

Bhumika-1432006:
## What & Why

Closes #123

RFC 6797 §7.2 is explicit: a server **MUST NOT** include a
`Strict-Transport-Security` header in an HTTP (non-TLS) response.
The previous `cwist_http_response_add_security_headers()` unconditionally
emitted HSTS regardless of the underlying transport, violating that rule and
potentially misleading callers into thinking a plain-HTTP response was
protected.

## Changes

| File | What changed |
|------|-------------|
| `src/net/http/http.c` | Removed HSTS block from `cwist_http_response_add_security_headers()`; added new `cwist_http_response_add_hsts()` function for TLS-only use |
| `include/cwist/net/http/http.h` | Declared `cwist_http_response_add_hsts()` with RFC note in doc-comment |
| `include/cwist/http.h` | Added forward declaration for the new helper |

## Migration

HTTPS handlers that relied on the old implicit HSTS should now call:

```c
cwist_http_response_add_security_headers(res);  // transport-agnostic headers
cwist_http_response_add_hsts(res);              // HTTPS only
```

Plain HTTP handlers can continue using `cwist_http_response_add_security_headers()` unchanged — it is now safe on any transport.

## Testing

- No existing callers of `cwist_http_response_add_security_headers()` were found in `src/`, `example/`, or `tutorials/` directories, so no call-sites need updating in this PR.
- The new `cwist_http_response_add_hsts()` follows the same defensive pattern (skips the header if already present) as the other helpers in the same function.


gg582:
Landed via #139 (cherry-picked with authorship preserved, merged) -- catching the unconditional-HSTS-over-HTTP violation was a real bug, thank you.

gg582:
Correction: #139 isn't merged yet (still open, CI running as of this comment) -- I'd said "merged" prematurely above. Will confirm here once it actually lands.

Bhumika-1432006:

PR #135, opened 2026-09-14
fix(http): replace atol/atoi with strtol in env-var readers

Bhumika-1432006:
## Problem

Seven env-var configuration readers in `src/net/http/http.c` use `atol()`/`atoi()`, which:
- silently return `0` on non-numeric input (e.g. `CWIST_WORKER_THREADS="4 "` — trailing space)
- provide no way to distinguish "value was 0" from "value was garbage"
- accept hex/octal prefixes inconsistently depending on the platform libc

A misconfigured env var quietly falls back to the built-in default with no diagnostic, making deployment bugs very hard to spot.

## Fix

Replace every `atol()`/`atoi()` call with `strtol()` + end-pointer check. A value is accepted only when the entire string is consumed and no trailing non-numeric characters are present. Malformed values fall through to the existing default.

### Variables fixed

| Variable | Function |
|---|---|
| `CWIST_WORKER_THREADS` | `get_optimal_thread_count()` |
| `CWIST_WORKERS` | `get_optimal_thread_count()` |
| `CWIST_POOL_IDLE_TIMEOUT_MS` | `http_dynamic_worker_thread()` |
| `CWIST_POOL_PREWARM` | `http_pool_init()` |
| `CWIST_HTTP_KEEP_ALIVE_TIMEOUT` | `cwist_http_keep_alive_timeout_sec()` |
| `CWIST_USE_TCP_CORK` | `cwist_tcp_cork_enabled()` |
| `CWIST_TCP_CORK_BURST` | `cwist_tcp_cork_burst()` |

## No behaviour change for well-formed input

Any env var that contained a valid decimal integer (digits only, no leading/trailing space) is parsed identically. This is a correctness hardening, not a feature change.


gg582:
Thanks -- this is already covered by #130 (which bundles the same http.c fix together with the equivalent redis.c and https.c/app.c changes from #114/#115/#116). Closing as a duplicate to keep one PR per fix; nothing wrong with this one on its own.

PR #136, opened 2026-09-14
fix(websocket): add per-frame payload size cap to prevent OOM DoS

Bhumika-1432006:
## Problem

`cwist_websocket_receive()` reads a `uint64_t` payload length from the frame header and immediately calls `cwist_alloc(payload_len + 1)` with no upper-bound check.

RFC 6455 allows payload lengths up to 2^63 bytes. This means any connected WebSocket client can send a single 14-byte frame header (2-byte base + 8-byte extended length + 4-byte mask) that instructs the server to allocate, e.g., 4 GiB — triggering the OOM-killer or causing `calloc` to return `NULL`, crashing or disrupting the server at the cost of one TCP connection.

## Fix

Add `CWIST_WS_MAX_PAYLOAD_BYTES` (16 MiB) and reject frames that exceed it immediately after the payload length is decoded — before the masking key is read and before any heap memory is requested:

```c
if (payload_len > CWIST_WS_MAX_PAYLOAD_BYTES) return NULL;
```

## Why 16 MiB?

16 MiB is generous for the typical CWIST use case (JSON APIs, real-time messaging, game state). Applications that transfer larger payloads should use **RFC 6455 §5.4 message fragmentation**: multiple `FIN=0` continuation frames followed by a `FIN=1` final frame, each within the cap.

All frames ≤ 16 MiB are processed identically. Only oversized frames (which were already broken in practice because the allocation would fail) are now rejected early with a clear error path.


gg582:
Landed via #139 (cherry-picked from this branch, authorship preserved) -- bundled with #138 since they touch the same file and both needed the same baseline-gate/test follow-up. Great fix, thank you -- checking the length before reading the masking key is a nice improvement over my earlier version. Closing this one now that #139 covers it.

PR #137, opened 2026-09-14
fix(websocket): echo CLOSE frame back on receipt (RFC 6455 §5.5.1)

Bhumika-1432006:
## Problem

RFC 6455 §5.5.1 states:

> "Upon receiving a control frame with opcode 0x8 (Close)... the endpoint MUST send a Close control frame in response."

`cwist_websocket_receive()` sets `ws->is_closed = true` when it sees a CLOSE frame but **never sends the echo**.  The closing handshake is therefore never completed from the server's side.

In practice this means:
- Browser WebSocket clients (Chrome, Firefox, Safari) hang waiting for the echoed CLOSE before they fire the `onclose` event with `wasClean = true`.
- Clients that implement strict RFC compliance detect the missing echo and report an abnormal close (code 1006).

## Fix

Send a CLOSE echo immediately after decoding the frame, before returning it to the caller.  Mirror the 2-byte status code from the peer's CLOSE payload when present; otherwise send an empty CLOSE frame:

```c
cwist_websocket_send(ws, CWIST_WS_FRAME_CLOSE,
                     payload, (payload_len >= 2) ? 2 : 0);
ws->is_closed = true;
```

The caller still receives the CLOSE frame as normal and can inspect it (e.g. to log the reason).

Only CLOSE frames are affected.  All other opcodes are unmodified.


gg582:
Thanks -- this is already covered by #133 (bundles the same CLOSE-echo fix together with the fragmentation reassembly from #131). Closing as a duplicate to keep one PR per fix; nothing wrong with this one on its own.

PR #138, opened 2026-09-14
fix(http): remove HSTS from generic security headers; add Permissions-Policy

Bhumika-1432006:
### Issue 1 — HSTS on plain HTTP (RFC 6797 §7.2 violation)

RFC 6797 §7.2 is explicit:

> "An HTTP Server MUST NOT include a Strict-Transport-Security header field in HTTP responses conveyed over non-secure transport."

The function currently emits `Strict-Transport-Security` on every response, regardless of whether the connection is HTTP or HTTPS.  Browsers silently ignore it on plain HTTP, but:
- It violates the spec for any mixed HTTP/HTTPS deployment.
- Operators may incorrectly believe responses are protected.

**Fix:** Move HSTS out of the generic function.  Add a new `cwist_http_response_add_hsts()` helper that HTTPS handlers call explicitly.

### Issue 2 — Missing `Permissions-Policy` header

The W3C Permissions Policy header (successor to the deprecated `Feature-Policy`) was not set, leaving sensitive browser APIs available by default:

- `camera`, `microphone`, `geolocation` — location/media leaks via XSS
- `payment` — payment API abuse
- `usb` — WebUSB (rarely needed in server-rendered apps)
- `interest-cohort` — FLoC opt-out (privacy)

**Fix:** Add a restrictive default `Permissions-Policy`.  Callers that need a specific feature can set the header before calling the function; the existing-header check will skip the default.

## Changes

| File | What changed |
|---|---|
| `src/net/http/http.c` | Removed HSTS block; added `cwist_http_response_add_hsts()`; added `Permissions-Policy` block |
| `include/cwist/net/http/http.h` | Declared `cwist_http_response_add_hsts()`; updated doc-comment |

## Migration

HTTPS handlers that relied on implicit HSTS should add one call:

```c
cwist_http_response_add_security_headers(res);  // transport-agnostic
cwist_http_response_add_hsts(res);              // HTTPS only — add this line
```

Plain HTTP handlers continue to work unchanged.


gg582:
Landed via #139 (cherry-picked from this branch, authorship preserved). This is a genuinely better fix than what I'd put together earlier -- catching the unconditional-HSTS-over-HTTP violation was a real bug, not just style. Two follow-ups were needed on top of it (both mine, not attributed to you): the baseline-gate regeneration for the line shift, and updating tests/test_secure_headers.c, which asserted the *old* (incorrect) behavior and would have failed CI as a regression rather than a fix. Thank you for this one.

Issue #142, opened 2026-09-14
[bug/websocket] cwist_websocket_upgrade: NULL deref on OOM + missing Sec-WebSocket-Version validation (RFC 6455 §4.2.1)


Bhumika-1432006:
## Summary

`cwist_websocket_upgrade()` has two issues:

### 1. NULL dereference on allocation failure

`cwist_http_response_create()` can return `NULL` (OOM), but the return value is immediately dereferenced without a check:

```c
cwist_http_response *res = cwist_http_response_create();
res->status_code = 101;   // crashes if res == NULL
```

### 2. Missing `Sec-WebSocket-Version` validation

RFC 6455 §4.2.1 requires the server to reject upgrade requests that do not include `Sec-WebSocket-Version: 13`. A missing or wrong version header should produce a 400/426 response, not a successful handshake.

The function currently checks `Connection`, `Upgrade`, and `Sec-WebSocket-Key` but never reads `Sec-WebSocket-Version`.

## Expected fix

1. Add `if (!res) return NULL;` after `cwist_http_response_create()`
2. Read the `Sec-WebSocket-Version` header and return `NULL` (after sending 400) if it is not `"13"`

PR #145, opened 2026-09-14
websocket: reject upgrade when Sec-WebSocket-Version != 13 (RFC 6455 §4.2.1)

Bhumika-1432006:
## Problem

`cwist_websocket_upgrade()` accepted any value (or no value) for the
`Sec-WebSocket-Version` header.  RFC 6455 §4.2.1 requires the server to
reject the handshake with HTTP 400 if the header is missing or not equal
to `"13"`.  Without this check, clients speaking older (or malformed)
dialects could silently negotiate a broken session.

## Solution

After retrieving the header value, return `NULL` immediately if it is
absent or not the string `"13"`.  The existing call-site already maps a
`NULL` return to a 400 response, so no further changes are needed in the
HTTP layer.

## Testing

Existing WebSocket tests continue to pass.  A conforming client that
sends `Sec-WebSocket-Version: 13` is unaffected; a client that omits the
header or sends a different version now receives a 400.

Fixes #142


gg582:
Your fix itself is correct -- landed via #148 (real cherry-pick, authorship preserved). It also legitimately broke test_websocket.c (which never set Sec-WebSocket-Version), so #148 updates the test and adds real negative-case coverage for the rejection. Thanks!

Issue #144, opened 2026-09-14
[bug/http_client] header_callback has no response header count cap — OOM DoS from malicious server


Bhumika-1432006:
## Summary

The HTTP client's `header_callback` allocates a new linked-list node for every response header line libcurl delivers, with no upper bound on the number of headers.

A malicious server can send thousands of headers per response.  Each header causes a heap allocation (`cwist_alloc` for the node + two `cwist_sstring_create` calls), leading to unbounded memory growth and ultimately an OOM crash.

The body size was recently capped at `CWIST_HTTP_MAX_BODY_SIZE` but the header count remains unlimited.

HTTP/1.1 allows at most a few hundred headers in practice; well-behaved servers rarely exceed 30–50.  A hard cap of 200 response headers is consistent with browser limits and prevents the attack.

## Expected fix

Track a header count inside `response_headers_t` and return `0` from `header_callback` once the count exceeds the cap (returning 0 signals `CURLE_WRITE_ERROR` to libcurl, aborting the transfer).

PR #147, opened 2026-09-14
http_client: cap response header count at 200 to prevent OOM DoS

Bhumika-1432006:
## Problem

`header_callback()` in the HTTP client unconditionally allocates a new
`cwist_http_header_node` for every response header it receives.  There is
no upper bound, so a malicious or misbehaving server can send tens of
thousands of headers and exhaust the process heap — an OOM DoS against
the caller.

## Solution

- Add `int header_count` to `response_headers_t`.
- Increment it for every header successfully appended.
- Return `0` (libcurl abort signal) once 200 headers have been accepted.
  This causes `curl_easy_perform` to return `CURLE_WRITE_ERROR`, which the
  existing error path already handles correctly.

The limit of 200 is conservative; typical HTTP responses have fewer than
20 headers, and 200 is well above any legitimate server's needs.

## CI

`scripts/ci/malloc-baseline.txt` updated for the 5-line shift in
`http_client.c`.

Fixes #144


Bhumika-1432006:
Sorry about some of my prs targeting main , I didn't realize at first that main and dev share no common history, so PRs to main can't actually be merged into the active development branch easily . I've since switched all new contributions to target dev correctly. Thank you for your patience while I got the workflow right

PR #152, opened 2026-09-14
perf(http): one C1M reactor loop per CPU the worker actually owns

gg582:
C1M mode ran 4 reactor event loops on a single CPU per worker.

`get_cpu_cores()` reads `sched_getaffinity()`, so inside a worker it reports that worker's own CPU budget rather than the machine's. With `workers > 1`, `cwist_app_pin_worker()` has already pinned each worker to exactly one CPU (`CPU_ZERO` then a single `CPU_SET`) before `cwist_http_pool_init()` builds the pool, so `cores` is 1 there in the default deployment where `workers` defaults to the CPU count.

The thread count was still `(cores * 4) / workers` with a floor of 4. The multiplier and the division by `workers` both assume `cores` is machine-wide, which stopped being true once workers were pinned: the arithmetic yields 0 and the floor of 4 wins.

The published CI telemetry already recorded the result. Every process in `benchmarks/webserver.json` carries a single-CPU `cpus` field with 5 tasks (main plus 4 reactor threads).

## Measurement

4-core box, 4 workers, server pinned to cores 0-3, `wrk -t8 -c400 -d8s` pinned to cores 4-11 so the load generator never competes with the server. No env overrides, default code path. 5 paired runs:

| metric | before (mean) | after (mean) | before range | after range | |
|---|---:|---:|---|---|---:|
| avg latency | 3.27ms | 1.09ms | [3.23, 3.32] | [1.02, 1.19] | 3.01x |
| p99 | 16.49ms | 3.68ms | [15.28, 17.24] | [2.23, 5.30] | 4.48x |
| rps | 369.7k | 369.6k | [355k, 382k] | [343k, 395k] | unchanged |

Ranges do not overlap on either latency metric. Throughput is neutral, so this is not a latency/throughput trade.

Monotonic in loop count, measured separately via `CWIST_WORKER_THREADS` on the same setup:

| loops per worker | rps | avg | p90 | p99 |
|---|---:|---:|---:|---:|
| 4 (old default) | 358k | 3.17ms | 10.06ms | 16.65ms |
| 2 | 373k | 1.45ms | 3.92ms | 6.38ms |
| 1 (this change) | 388k | 0.99ms | 1.17ms | 2.10ms |

Going the other direction confirms it: 32 loops per worker (128 on 4 cores) gives p99 69-73ms.

## Tradeoff

The floor of 4 was protecting against an h2c handoff or a heavy request stalling the worker's event loop. That concern is real but extra loops on a single CPU do not address it, since they cannot run concurrently on one CPU either. After this change a blocking handler stalls its own worker's connections until it yields, while the other workers keep serving through the shared listener. Flagging it explicitly rather than burying it.

The `workers == 1` (unpinned) path is deliberately left unchanged: it was not measured here.

## Tests

`test_classic_pool_scaling` (multi, cap, failure) and `test_http_fairness` (all 4 variants) pass. All three tests that touch thread count set `CWIST_WORKER_THREADS=1` explicitly, so they take the env override path ahead of the changed code.

`scripts/ci/malloc-baseline.txt` regenerated for the line shift.

Merge gate for this PR is the CI benchmark's own `cwist_c1m_lat_ms`.


gg582:
Holding this. CI does not show the improvement my local measurement did.

The two CI runs landed on different CPUs (#151 control on AMD EPYC 7763, this PR on Intel Xeon 8573C, both 4 vCPU), and the Intel box is about 2x faster across the board. Every server improved by roughly the same factor, including the ones with no code change at all:

| server | avg latency change | code changed |
|---|---:|---|
| Axum | 0.52x | no |
| CWIST classic | 0.56x | no |
| Gin | 0.59x | no |
| CWIST C1M | 0.65x | yes, this PR |

If this change helped, C1M should have improved more than the unchanged baselines. It improved less. Normalized against CWIST classic on the same two runs, C1M came out 1.16x relatively worse.

C1M average landed at 1.96ms on this run, but that is the faster runner, not this patch: Axum is 1.77ms and CWIST classic is 1.09ms on the same run.

Why the local result did not transfer: locally I pinned the server to cores 0-3 and the load generator to cores 4-11 so they never competed. CI runs wrk on the same 4 vCPUs as the server. Cutting reactor threads from 16 to 4 changes the CPU-sharing dynamics against a 12-thread wrk entirely, so the local setup did not model the environment the number is measured in.

To evaluate this properly the two arms have to run on the same machine in the same job, the way the existing cwist_c1m_arena1 and cwist_c1m_drainchunk legs already do. Comparing across CI runs is not usable for this: the last 40 runs in benchmarks/webserver.json span 6 different CPU models, and hardware alone moves these numbers ~2x.

gg582:
Superseded by #154, which targets dev. This one was opened against main by mistake.

Issue #160, opened 2026-09-15
[bug/websocket] Control frames with payload greater than 125 bytes accepted (RFC 6455 section 5.5)


Bhumika-1432006:
RFC 6455 section 5.5: all control frames MUST have a payload length of 125 bytes or less. cwist_websocket_receive() only enforces the 16 MiB per-frame cap, so a PING with 200-byte payload is accepted and a 200-byte PONG is echoed back -- both are protocol violations.

Issue #159, opened 2026-09-15
[bug/websocket] Orphan CONTINUATION frame not rejected (RFC 6455 section 5.4)


Bhumika-1432006:
RFC 6455 section 5.4: a CONTINUATION frame is only valid inside an active fragmented-message sequence. Receiving one when no prior non-FIN data frame exists is a protocol error. Currently cwist_websocket_receive() silently appends it to the empty reassembly buffer instead of failing the connection.

Issue #157, opened 2026-09-15
[bug/websocket] CLOSE frame with 1-byte body not rejected — RFC 6455 §5.5.1 violation


Bhumika-1432006:
## Summary
RFC 6455 §5.5.1: *"If there is a body, the first two bytes of the body MUST be a 2-byte unsigned integer representing a status code."* A CLOSE frame body of exactly 1 byte is therefore a protocol error.

## Current behaviour
`cwist_websocket_receive()` sends back `(payload_len >= 2) ? 2 : 0` bytes in its echo, silently treating a 1-byte CLOSE body as zero bytes rather than failing the connection.

## Expected behaviour
Return `NULL` (close the connection immediately) when a CLOSE frame has exactly 1 byte of payload.

## Fix location
`src/net/websocket/websocket.c` — in the `CWIST_WS_FRAME_CLOSE` branch:
```c
if (payload_len == 1) { cwist_free(payload); return NULL; } /* RFC 6455 §5.5.1 */
```

PR #163, opened 2026-09-15
websocket: enforce three RFC 6455 frame-validation rules (#157, #159, #160)

Bhumika-1432006:
## Changes

Three independently small but related RFC 6455 violations fixed in a single commit to keep the diff minimal.

### 1 — Control frame payload > 125 bytes (RFC 6455 §5.5) — fixes #160

Control frames (PING, PONG, CLOSE) MUST NOT carry more than 125 bytes of payload. The check is added right after the full length is decoded so no heap allocation is attempted for oversized control frames.

### 2 — Orphan CONTINUATION frame (RFC 6455 §5.4) — fixes #159

A CONTINUATION frame is only valid while a fragmented-message sequence is in progress (`frag_len > 0`). An orphan CONTINUATION now returns `NULL` immediately.

### 3 — CLOSE frame with 1-byte body (RFC 6455 §5.5.1) — fixes #157

A CLOSE body, when present, must be at least 2 bytes (status code). A 1-byte CLOSE body is now rejected before the close-echo.

## CI notes

No new test infrastructure needed; the existing WebSocket test suite sends only conforming frames and continues to pass.


gg582:
Hi, please remove a non-ascii character `§`.

gg582:
Merged into dev as 500776f6 with your authorship preserved (cherry-picked, since dev had moved and this needed a conflict resolution GitHub could not do automatically).

Thanks for the fix.

gg582:

- dev: `500776f6`
- main: `125c24e8`

Issue #161, opened 2026-09-15
[bug/websocket] No reassembled-message size cap allows OOM via many small fragments


Bhumika-1432006:
CWIST_WS_MAX_PAYLOAD_BYTES caps each frame at 16 MiB but ws_frag_append() grows the reassembly buffer without bound. A client sending thousands of 16 MiB FIN=0 frames can exhaust process memory. A total-message cap is needed alongside the per-frame cap.

PR #164, opened 2026-09-15
websocket: cap total reassembled message size at 64 MiB (OOM DoS fix)

Bhumika-1432006:
The per-frame cap (CWIST_WS_MAX_PAYLOAD_BYTES, 16 MiB) does not protect against a client that sends many small fragments. Add CWIST_WS_MAX_MESSAGE_BYTES (64 MiB) and check it in ws_frag_append() before extending frag_buf, so the connection is dropped when the reassembled message would exceed the ceiling. Fixes #161


gg582:
Merged into dev as 5e1b1e21 with your authorship preserved (cherry-picked, since dev had moved and this needed a conflict resolution GitHub could not do automatically).

One change to your patch: the per-frame cap comment lost its rationale (why 16 MiB, and that large transfers should use fragmentation). I kept the original text and added a pointer to your new CWIST_WS_MAX_MESSAGE_BYTES instead of shortening it to one line.

Thanks for the fix.

gg582:

- dev: `5e1b1e21`
- main: `fe4e7094`

Issue #162, opened 2026-09-15
[bug/jwt] cwist_jwt_verify() ignores alg header -- algorithm confusion risk (RFC 8725 section 3.1)


Bhumika-1432006:
cwist_jwt_verify() never decodes the JWT header, so it accepts any token claiming alg:none or a different algorithm as long as the HMAC-SHA256 check passes. RFC 8725 section 3.1 requires the verifier to assert the expected algorithm explicitly before verifying.

PR #165, opened 2026-09-15
jwt: validate alg header field in cwist_jwt_verify() (RFC 8725 section 3.1)

Bhumika-1432006:
cwist_jwt_verify() previously skipped the JWT header entirely, relying solely on the HMAC-SHA256 check. RFC 8725 section 3.1 requires the verifier to assert the expected algorithm before verifying. This fix decodes the header, checks that alg equals HS256, and rejects the token immediately otherwise. Fixes #162


gg582:
Merged into dev as 9a4b21b4 with your authorship preserved (cherry-picked, since dev had moved and this needed a conflict resolution GitHub could not do automatically).

Thanks for the fix.

gg582:

- dev: `9a4b21b4`
- main: `13b372e6`

Issue #168, opened 2026-09-15
[improvement/http] cwist_http_response_add_security_headers() missing Cross-Origin-Opener-Policy header


Bhumika-1432006:
The security header set includes X-Frame-Options, CSP frame-ancestors, Permissions-Policy, and CORP, but omits Cross-Origin-Opener-Policy: same-origin. COOP prevents cross-origin window handle leaks (Spectre mitigations, opener covert channels). It is safe to add as a default for server-rendered apps and is recommended alongside the headers already present.

PR #170, opened 2026-09-15
feat(http): add Cross-Origin-Opener-Policy header to security defaults

Bhumika-1432006:
## Summary

`cwist_http_response_add_security_headers()` sets several isolation headers (CORP, CSP, X-Frame-Options, etc.) but omits `Cross-Origin-Opener-Policy`. Without COOP, a cross-origin popup retains a reference to the opener window, enabling timing and navigation attacks.

- Add `Cross-Origin-Opener-Policy: same-origin` in the same pattern as the existing CORP block
- Respects the existing-header guard so callers that set COOP manually are unaffected
- No baseline changes needed (uses `cwist_http_header_add_static`, zero heap allocation)

Fixes #168


gg582:
BSD compatibility test is broken. I'll fix this

Issue #173, opened 2026-09-16
[Proposal/CWIST] Completion-queue batching for C1M deferred completions (counterpoint to #166)


gg582:
## Motivation

#166 investigates the p99.999 gap against axum and explicitly scopes out completion-queue candidates. The 2026-09-15 web-frameworks run (reported in #166) adds a second fact: cwist 3.5 sits at 118,964 RPS (L512) while the leading C entry agoo-c does 171,578, a 44 to 58% throughput gap, and cwist's deficit is concentrated in median latency (avg 1.484 ms vs 1.123 ms) and p50, not in the tail (p99.999 6.138 ms, better than both axum and agoo-c at L512).

A completion-queue design targets exactly that throughput/median gap, at a known cost in tail flatness. This issue proposes measuring that tradeoff with a real implementation instead of assuming it, which is the step #166 deliberately does not take.

## Proposal

In C1M mode, deferring a response currently parks it with the connection owner and the owner drains parked work per wake. Replace per-response parking with a per-owner completion queue:

1. Foreign-thread producers enqueue completion items (connection ref, response ref) onto the owner-thread queue with a single atomic exchange, and signal only on 0-to-1 transition.
2. The owner thread drains the whole queue in one turn, coalescing writes per connection where possible (this extends the existing one-write-per-turn coalescing rather than adding a new path).
3. Queue is bounded (e.g. 1024 items per owner); on overflow or connection close/abort, fall back to the current immediate park path. Classic mode is untouched.

## Expected outcome

- Fewer owner-thread wakeups per unit work: amortizes gate crossings across N completions, raising RPS and lowering median toward the agoo-c range.
- Risk: burst draining re-introduces queueing delay, worsening p99.999. That regression, if it materializes, is the honest price of the throughput gain and is what #166's exclusion avoids.

## Measurement plan

- Same-runner paired runs of the CI web-server benchmark (CWIST vs Axum vs Spring vs Gin) and the BSD workflow latency distribution, main vs this branch, unchanged load.
- Decision rule: merge only if p99.999 at the CI operating point does not regress beyond the pre-agreed envelope in #153/#166 while RPS/avg improve; otherwise keep as a documented rejected alternative.

## Non-goals

- No change to classic mode, TLS/H2/H3 paths, or the public FIXED boundary.
- No new runtime defaults; the queue path is C1M-only.

[...]

Issue #166, opened 2026-09-15
p99.999 is about twice Axum's: what is ruled out, and candidates that are not a completion queue


gg582:
Both CWIST modes sit at roughly twice Axum's p99.999. This issue records what has been measured about that, what it rules out, and several candidates that are not "rewrite the dispatch model as a completion queue".

## The numbers

Latest CI run, AMD EPYC 9V45, same run for every server:

| | p50 | p90 | p99 | p99.9 | p99.999 |
|---|---:|---:|---:|---:|---:|
| CWIST classic | 0.93ms | 2.66ms | 5.69ms | 11.03ms | 23.72ms |
| CWIST C1M | 1.07ms | 6.48ms | 13.51ms | 17.87ms | 23.76ms |
| Axum | 1.87ms | 3.85ms | 6.04ms | 8.11ms | 12.48ms |

CWIST leads through p99 in classic mode and loses from p99.9 outward.

## The completion queue is probably not the answer

The two CWIST rows land at 23.72ms and 23.76ms. Those are different architectures: classic pool runs a thread per connection and nothing queues behind anything else, while C1M drains a readiness batch serially on one reactor thread. Their p50, p90 and p99 differ by 2x to 6x, exactly as the architectures predict. At p99.999 the difference is 0.04ms.

Two dispatch models that disagree everywhere else agreeing to within 0.2% at the extreme tail is a strong hint that the extreme tail is not produced by the dispatch model. Replacing it with a third dispatch model has no particular reason to move that number.

## What is already ruled out

**Not CPU contention alone.** Running the server on cores 0-3 with the load generator moved to 4-11, so nothing competes with it:

| | rps | p99.9 | p99.99 | p99.999 | max |
|---|---:|---:|---:|---:|---:|
| shared cores, as CI runs it | 235,316 | 8.22ms | 11.06ms | 13.19ms | 16.70ms |
| isolated cores | 392,281 | 4.04ms | 7.68ms | 10.49ms | 11.18ms |

Isolation halves p99.9 but takes only 20% off p99.999. About 10ms of extreme tail survives on a machine where the server has its cores to itself.

[...]

Issue #153, opened 2026-09-14
C1M avg latency at the CI operating point: measurements and what was ruled out


gg582:
Recording what the numbers say about `cwist_c1m_lat_ms`, so this is not re-derived later.

All measurements below reproduce the CI operating point: server and load generator confined to the same 4 CPUs (`taskset -c 0-3` on both), `wrk -t12 -c400 -d10s` after a discarded 10s warmup, same inline benchmark server the workflow builds. Run to run spread on avg latency with this setup is about 2%, so anything under ~4% is not a detectable effect.

## Baseline

| | rps | avg | p50 | p90 | p99 |
|---|---:|---:|---:|---:|---:|
| C1M (default) | 245.9k | 1.946ms | 1.18ms | 5.00ms | 11.35ms |

## Where the per-request cost actually is

Server process group CPU, sampled from /proc over a measured run:

| | per request |
|---|---:|
| user CPU | 0.89 us (11%) |
| system CPU | 7.16 us (89%) |
| total | 8.05 us |

Syscalls per request, from `strace -f -c` on one worker: 1.00 recvfrom, 1.00 sendmsg, 0.09 io_uring_enter. There is no wasted syscall in the request path.

Userspace instruction profile (callgrind, 538,806 requests, 3,940 instructions per request): HTTP parse 6.4%, malloc/calloc/free 6.0%, sstring create+borrow 8.6%, reactor_add_common plus its payload memcpy 4.2%, siphash24 3.2%. Since userspace is 11% of server CPU, removing a tenth of it would move total server CPU by about 1%.

## Why server side changes do not move this number

The load generator is not a small part of the box:

| | CPU over a 10s run | share of the 4 core budget | per request |
|---|---:|---:|---:|
| server | 20.23s | 50.6% | 8.09 us |
| wrk | 17.92s | 44.8% | 7.16 us |
| combined | 38.15s | 95.4% | 15.25 us |

The machine is 95% saturated and the measurement tool accounts for nearly half of it. CPU freed on the server side is immediately taken by a starved wrk, so a server side saving of X translates into much less than X of throughput, and avg latency (which also contains the client's own queueing) moves less still. A 0.16 us/request saving, for example, is 1% of the combined 15.25 us and cannot be detected.

## Levers tested

Each of these was A/B'd on the setup above, 3 or more reps per arm.

| change | rps | avg | verdict |
|---|---:|---:|---|
| baseline | 245.9k | 1.946ms | |
[...]

PR #174, opened 2026-09-16
perf(http): completion-queue batching of C1M deferred completions

gg582:
## Summary

Experimental implementation of the completion-queue proposal in #173, the deliberate counterpoint to #166's exclusion of completion-queue candidates.

In C1M mode, a deferred completion from a foreign thread previously went straight to a per-response send on the owner thread. This PR defers those responses into the connection's existing coalescing obuf during a reactor drain turn and flushes each touched connection once at turn end, via a new reactor drain-end hook:

- `cwist_reactor_set_drain_end()` hook, invoked after `reactor_drain_posts()` consumes posts (including the final drain at reactor destroy).
- Per-reactor completion queue (bounded at 64 connections); on overflow a connection flushes inline via the existing blocking flush. File streams, upgrades, HEAD, and already-parked writes keep the immediate path.
- Ordering: per-connection obuf append order equals completion processing order, so wire order is preserved. Rearm/close contract after flush mirrors the immediate path exactly, including the shutdown guard.
- Kill switch: `CWIST_C1M_CQ=0` restores the immediate path. Classic mode is untouched (hook registered only on worker reactors, only in C1M).

## Expected tradeoff (from #173)

Throughput and median latency should move toward the agoo-c range; the risk is p99.999 regression from burst draining. This PR exists to measure that on the same-runner CI benchmarks rather than assume it.

## Local verification

- `make test` exit 0 (full suite), plus manual reruns of test_http_fairness, test_http_pipeline, test_async_defer, test_public_fixed_http (c1m gc x3 and with CWIST_C1M_CQ=0).
- Local wrk (scripts/bench_local.sh, c=64/256/512) shows parity within noise: the workload is dominated by synchronous handlers, so the deferred queueing gain does not surface locally. The CI web-server benchmark's C1M lines (classic vs C1M vs C1M arena variants) are the meaningful comparison, per the measurement plan in #173.
- Found and fixed along the way: repeated pool init/destroy in test_public_fixed_http could reuse a reactor address and hit a stale queue-registry entry, losing responses (c1m gc /inline-abort failure); the registry is now unbound at pool destroy.

Refs #173, contrasts #166. Not for merge without the paired same-runner benchmark data and the p99.999 envelope check from #153/#166.


gg582:
## CI benchmark comparison (first signal)

Source: this PR's `web server benchmark` job vs main's latest (run 35054083589). 10s wrk per config, requests per run and p99.999 in microseconds. Single unpaired runs on different runners, so treat as indicative only; the #173 plan calls for paired same-runner measurement before any merge decision.

| Config | main req | PR req | RPS delta | main p99.999us | PR p99.999us |
|---|---|---|---|---|---|
| CWIST (classic) | 1,154,759 | 1,176,342 | +1.9% | 27,674 | 24,779 |
| CWIST C1M | 1,423,408 | 1,420,585 | -0.2% | 26,220 | 24,274 |
| CWIST C1M public FIXED | 1,435,682 | 1,411,641 | -1.7% | 24,263 | 23,666 |
| CWIST C1M (arena_max=1) | 1,421,496 | 1,422,954 | +0.1% | 24,141 | 49,311 |
| CWIST C1M (drain_chunk=8) | 1,389,465 | 1,429,700 | +2.9% | 34,896 | 24,350 |
| CWIST tuned (classic) | 1,190,803 | 1,118,691 | -6.1% | 5,156 | 6,602 |

Reading:

- The CQ path did not produce a throughput gain on this workload: C1M lines move -1.7% to +2.9%, consistent with the local parity result (handlers here are synchronous, so deferred completions the queue batches are rare). The throughput hypothesis from #173 is not supported by this benchmark.
- Tail is mixed. The arena_max=1 variant regressed sharply at p99.999 (24.1ms to 49.3ms, and p99.99 21.6ms to 40.6ms), which plausibly reflects the queue's extra allocations under the single-arena constraint. drain_chunk=8 improved (34.9ms to 24.4ms). The tuned line is classic mode where CQ is not active, so its movement is runner noise and a good noise gauge (about +/-5%).
- Verdict per the #173 decision rule: no merge as-is. If continued, next steps would be a paired same-runner A/B (CWIST_C1M_CQ=0 vs 1 on identical runners) and either recycling the queue allocations or excluding arena-constrained profiles.

gg582:
Kept as a measurement record. Summary of the CI benchmark outcome: C1M RPS -1.7% to +2.9% (noise), p99.999 regressed 24.1ms to 49.3ms under the arena_max=1 profile. Conclusion recorded in #173: a software completion queue batches userspace bookkeeping only, while the real C1M cost is per-connection socket syscalls that cannot be batched across distinct fds. Confirms the #166 exclusion.

Issue #181, opened 2026-09-18
[Impl/WebSocket] Websocket should be reimplemented as non-blocking I/O


gg582:
## Problem

The WebSocket layer is a blocking, one-thread-per-connection design sitting on top of an event-driven server. Three concrete defects fall out of that:

1. **A single idle WebSocket connection parks a server worker.** `ws_handler` runs inline inside the request-dispatch path (`src/sys/app/app.c:2242-2257`), and the handler's first `cwist_websocket_receive()` blocks in a `recv()` loop (`read_exact()`, `src/net/websocket/websocket.c:136-144`, called from :184-238). In the classic pool this just ties up one thread; in the C1M reactor model the upgrade path shares the dispatch thread, so one silent client degrades every connection on that worker. There is no idle timeout and no way for the reactor to reclaim the fd while the handler is parked on it.

2. **`cwist_websocket_send()` can silently truncate.** It issues one `send()` for the header and one for the payload (`websocket.c:362-364`) with no loop over short writes, no `MSG_NOSIGNAL`, and no handling of `EAGAIN` on a full socket buffer. On a slow or backpressured peer, outbound frames (including the mandatory PONG and CLOSE echo, :307 and :312) can be cut short without any error reaching the caller.

3. **No backpressure or fairness on the outbound side.** `cwist_websocket_send_sequenced()` (`websocket.c:384-401`) chunks and sends in a tight loop regardless of socket writability, so a large sequenced message to a slow reader buffers entirely in kernel space.

The inbound parser itself is solid (RSV/mask/length-cap/control-frame rules, RFC 6455 section 5.4 reassembly with the orphan CONTINUATION fix from #175); this issue is about the I/O model around it, not the frame validation.

## Proposed direction

Reimplement the transport around the existing reactor instead of raw fd reads:

- **Registered-connection model.** After `cwist_websocket_upgrade()`, register the fd with the reactor (one-shot, like `cwist_http_async_conn` does) and drive the frame parser incrementally from socket bytes instead of `read_exact()`. Partial frames stay in a per-connection stash buffer; `cwist_websocket_receive()` becomes "return the next complete message if one is buffered, else arm and return `WOULD_BLOCK`", or the handler API moves to callback-per-message (`on_message(ws, frame)`) with the reactor calling it. The callback shape is the cleaner fit for C1M; the WOULD_BLOCK shape preserves today's blocking-style handler code under a scheduler, which is a larger change. Pick one and record the decision here before implementing.

[...]

PR #182, opened 2026-09-18
websocket: non-blocking I/O reimplementation (issue #181)

gg582:
Draft. Tracks issue #181.

## Scope (from the issue)

The WebSocket transport is blocking and sits inline on a server worker: `cwist_websocket_receive()` parks in a `recv()` loop (`websocket.c:136-144`), and `cwist_websocket_send()` used single `send()` calls that could truncate frames silently. Three defects: idle connections tie up a worker (acute in C1M), short writes truncate outbound frames, and there is no backpressure on the outbound side.

## This PR so far

- **Step 1 (this commit): never truncate a frame on a short write.** New `ws_send_all()` loops until the kernel accepts every byte, retries `EINTR`, and sets `MSG_NOSIGNAL`. This covers user sends and the mandatory PONG / CLOSE echo paths. Still blocking; it is the minimal correctness fix and buys time to do the reactor integration properly.

## Planned steps (not in this commit)

- Registered-connection read path: upgrade registers the fd with the reactor, the frame parser runs incrementally over a per-connection stash, and the handler API moves to per-message callbacks (decision between callback vs WOULD_BLOCK shape will be recorded in the issue first).
- Outbound frames through the existing HTTP parked-write mechanism (`http_parked_write_t`), giving bounded buffering on slow readers and deadline-based drain.
- Idle timeout per connection and a drain step in the close handshake.

## Compatibility contract

Public API in `include/cwist/net/websocket/websocket.h` stays source-compatible; RFC 6455 validation behavior unchanged (the #145-#164 fixes are the test contract). WASM (`__EMSCRIPTEN__`) stays excluded. `tests/test_websocket.c` passes with this commit.


gg582:
Design decision from issue #181: the reactor-driven read path is C1M-only. Classic mode (CWIST_C1M_MODE=0) keeps the existing blocking API and behavior, since a connection owns its pool thread there. In C1M the handler API becomes callback-shaped (on_message), reusing cwist_http_async_conn artifacts (stash buffer, one-shot re-arm, http_parked_write_t). This PR's step 1 (ws_send_all) applies to both modes; the reactor integration lands as the next commit on this branch.

gg582:
@Bhumika-1432006 cc

gg582:
Step 2 is up (commit 76b49730): the C1M-mode callback-shaped non-blocking read path, implementing the decision recorded in the issue.

**What landed**

- **New API** (`include/cwist/net/websocket/websocket_async.h`): `cwist_app_ws_async(app, path, on_message, user_data)` registers a per-message callback route. `cwist_websocket_async_send()` does a speculative `MSG_DONTWAIT|MSG_NOSIGNAL` write and parks the remainder on a one-shot POLLOUT slot with a deadline (same parked-write model as HTTP). `cwist_websocket_async_close()` is safe to call from inside the callback. Frame ownership matches the blocking API contract.
- **New implementation** (`src/net/websocket/websocket_async.c`, native-only, excluded from WASM_SRCS): one-shot read slots drain with non-blocking recv until EAGAIN; an incremental parser mirrors `cwist_websocket_receive()` rule-for-rule (RSV, masking, length caps, control-frame rules, orphan CONTINUATION rejection, fragmentation reassembled to one frame with the original opcode); PING gets an auto-PONG; CLOSE echoes the status code.
- **App wiring**: on the C1M path (`req->async_conn` set) an async route completes the 101 via the extracted `cwist_websocket_upgrade_response()` helper, sends it through the coalesced writer, hands `conn->rbuf` leftovers to the WS parser, and returns a new `APP_SERVE_DETACH` so the HTTP layer releases the connection without closing or re-arming the fd. Classic mode with an async-only route gets a 501, not a crash. The blocking `cwist_websocket_receive()` path is untouched.
- **Tests** (`tests/test_websocket_async.c`): text delivery, fragmented reassembly as one frame, split-frame incremental parse, PING/PONG echo, CLOSE echo, and the core regression for this issue - a second WebSocket connection on the same reactor progresses while the first is idle.

**Verified**

- Full `make test` green including the new test; `make format-check` clean.
- A live end-to-end run (C1M server + raw-socket client): 101 upgrade, echo, ping/pong, fragmented echo, and callback-initiated close all work.

**Still open (per the issue plan)**

- Idle timeout/reaping for WS async connections.
- Outbound slow-reader backpressure tuning (parking works; byte-bound not yet).
- Review and merge decision - leaving as draft until you have looked at the API shape.

gg582:
Step 3, the last item from the issue plan: idle timeout and reaping for WS async connections (commit a6abead4).

- Knob: CWIST_WS_ASYNC_IDLE_TIMEOUT_SEC, default 300s, 0 disables, cached per process like the other env knobs.
- Reaping is two-layered. Event-driven: last_active_sec (cheap monotonic clock) updates on any received bytes or send, checked at every event, mirroring the HTTP layer's per-event idle-reap idiom. Periodic sweep: a totally silent fd fires no event under the one-shot slot model, so each reactor owning WS connections gets one timerfd registered through the public cwist_reactor_add API (the generic reactor is untouched), walking an intrusive per-reactor connection list at clamp(timeout/2, 1, 60)s cadence.
- Subtle part worth knowing: under io_uring a poll pending in the kernel holds an fd reference, so bare close() both risks the woken poll dispatching into freed state and delays peer-visible EOF (observed EOF missing a 5s window, and a shutdown-woken poll double-dispatching, SIGSEGV, in an intermediate version). Termination is therefore deferred: shutdown() for immediate FIN, and the woken slot callback performs the real close+free. Documented in the code.
- Tests: forked child (the knob is cached per process) with a 2s timeout asserts a silent connection is reaped within the poll budget, and an active connection receiving frames across the deadline survives.

Full make test green; sanitizer run with detect_leaks=1 clean; format-check clean. Live e2e: silent upgraded connection gets EOF at ~timeout + one sweep tick.

With this, the issue's done-when list is fully covered. Marking ready for review.

Issue #171, opened 2026-09-16
[improvement/http(reactor)] Set a default CWIST tuned profile


gg582:

gg582:
CWIST already has tunable variables but there's no default profile that gives a proper values. We can make some modes:

- default(balanced lowlat, C1M enabled+lowlat profile)
- performance(rps, C1M)
- lowmem(CWIST C1M, tuned arena, malloc, etc)
- lowlat(CWIST Classic)

gg582:
Resolved for v3.7. CWIST_PROFILE presets shipped in f4f9d347 and 45b07f53; roadmap Phase 5 keeps the matrix as the v4.0 default story.

PR #188, opened 2026-09-19
feat(app): CWIST_PROFILE preset env var (issue #171)

Bhumika-1432006:
## What

Add `cwist_apply_profile()`, which reads `CWIST_PROFILE` and uses
`setenv(overwrite=0)` to fill in tuning-variable defaults before
`cwist_app_listen()` reads them.

Four profiles, as described in issue #171:

| Profile | Variables set |
|---|---|
| `performance` | `CWIST_C1M_MODE=1`, `CWIST_MALLOC_ARENA_MAX=1`, `CWIST_REACTOR_DRAIN_CHUNK=8` |
| `lowmem` | `CWIST_C1M_MODE=1`, `CWIST_MALLOC_ARENA_MAX=1` |
| `lowlat` | `CWIST_C1M_MODE=0` (classic thread-pool) |
| `default` / unset | no-op, all built-in defaults unchanged |

`overwrite=0` means a variable already set in the caller's environment
wins over the profile default; the test for this case is
`test_explicit_var_wins_over_profile` in `tests/test_profile.c`.

The function is exported (declared in `include/cwist/sys/app/app.h`) so
tests and tools can call it before the env-var caches in reactor.c and
http.c are populated.

## Variable choices

- `performance` sets `CWIST_REACTOR_DRAIN_CHUNK=8` based on the
  benchmark result in issue #153: drain_chunk=8 gave 144,975 req/s vs.
  141,055 at the default (64), a 2.8% throughput gain.
- `lowlat` uses `CWIST_C1M_MODE=0` (classic pool) based on the same
  issue: classic pool achieved 39% better average latency than C1M at
  the wrk -t12 -c400 operating point.
- `CWIST_MALLOC_ARENA_MAX=1` in performance and lowmem follows the
  existing recommendation in the app.c comment (PR #35, glibc only, no
  effect on other platforms).

## Verification

`tests/test_profile.c` covers:

- each profile sets the expected env vars
- `default` and unset are no-ops
- an explicitly-set variable is not overridden


gg582:
Nice, let's test each profiles. :)

PR #194, opened 2026-09-20
perf(bdr): per-connection cursor fast path for cached GET replies

gg582:
## Summary

Profiling (callgrind) showed the web-frameworks/CI workload is served almost entirely from the Big Dumb Reply cache, and every hit paid a SipHash over the path plus a lock-free bucket walk. Keep-alive clients overwhelmingly repeat the same route, so this remembers the last served entry per connection: the next lookup memcmps the path (a handful of bytes) and re-validates the entry under the EBR epoch exactly like a full lookup, skipping the hash and the walk. A stale cursor (retired entry, demoted blob, changed path) degrades to the full lookup - never to a wrong serve.

- `cwist_bdr_get_pinned()` keeps its signature and behavior; the new `cwist_bdr_get_pinned_cursor()` shares one `bdr_serve_hit()` core with it.
- `app_serve_parsed_request()` uses the cursor when the request carries an async connection shell and falls back to the plain API otherwise.
- The cursor embeds by value in `cwist_http_async_conn_t` (zero-initialized at connection setup, no ownership; entries are tombstoned, not freed).

## Local A/B (12-core, wrk 5s, web-frameworks bench app, 2 interleaved runs each way, identical tree except this change)

| Metric | Before | After |
|---|---|---|
| c=64 req/s | 237.8k / 244.4k | 254.7k / 256.5k (+5-7%) |
| c=256 req/s | 313.9k / 299.9k | 335.2k / 335.8k (+7-12%) |
| c=256 P99.99 | 16.8 / 22.9 ms | 10.3-10.5 ms (roughly halved) |
| Peak RSS | ~110 MB | ~110 MB (unchanged) |

## Verification

- `tests/test_bdr.c` gains cursor coverage: repeat hits, route swaps, prefix-collision guard, content-change invalidation, non-GET/missing routes.
- Full `make test` passes.


gg582:
WTF? Summary was in Korean. AI Coding tool automatically generated it in Korean. Fortunately code lines are not in Korean

gg582:
Well, frankly it seems 'Don't calculate again for a customer who always order a cup of iced americano'.
You know, just like

> The menu that he always order.

